PR #5, opened 2021-01-06
implemented google sign in and 

Flou21:


Ekwav:
Minimal style changes necessary

PR #23, opened 2021-01-25
info icon implemented

matthias-luger:


Ekwav:
[i] should be on the left side. Its hard to notice on the right on bigger screens

PR #42, opened 2021-03-02
display bin as badge

### LINKED ISSUE #40: Bin should be displayed as badge
As BIN is a boolean attribute it makes more sense to display it next to the item count on the auctionDetails site.
Either the badge is there or not. The checkmark is rather iritating.

matthias-luger:
change color scheme
fixes #40


Ekwav:
fixes #40

PR #66, opened 2021-04-09
implemented dark mode

matthias-luger:


src/utils/Parser/APIResponseParser.tsx
```diff
@@ -152,6 +152,8 @@ export function parsePlayer(player: any): Player {
 }
 
 export function parseAuctionDetails(auctionDetails: any): AuctionDetails {
+    console.log(new Date(auctionDetails.end));
+    console.log(auctionDetails.end);
```
Ekwav, line 156: Looks like this is left over from the previous issiue. Please remove

src/components/NavBar/NavBar.tsx
```diff
@@ -9,13 +9,47 @@ function NavBar() {
...
+    const itemHoverStyle = {
+        background: isDarkMode ? "#272b30" : "#FFFFFF"
+    }
+
+    /**
+            --gray: #7a8288;
+    --gray-dark: #3a3f44;
+    --primary: #3a3f44;
+    --secondary: #7a8288;
+        */
```
Ekwav, line 43: dead code, please remove

Ekwav (changes requested):

Ekwav:
Also the nonactive buttons on playerDetails and itemPrices site currently are btn-ligth wich doesn't really fit into the darktheme.
I would suggest btn-secondary instead of btn-ligth, however I am unsure how that would fit into the ligth theme.

Ekwav (approved):

PR #73, opened 2021-04-10
dark mode primary and secondary button switch

matthias-luger:


Ekwav:
This change has not the targeted effect because there is still the class btn-ligth on the `label` element (wich is brighter)
The active button is still darker than the inactive button wich is the wrong way around

Ekwav:
image

PR #75, opened 2021-04-10
token error fixed

### LINKED ISSUE #71: JWT has expired handling
When not used for about a week the Google id JWT expires. 
This leads to the error response:
```
data: "{"Slug":"invalid_token","Message":"JWT has expired."}"
```
This should be handled by reseting the google login via deleting it.

matthias-luger:
fixes #71

### DIFF
--- a/src/components/GoogleSignIn/GoogleSignIn.tsx
+++ b/src/components/GoogleSignIn/GoogleSignIn.tsx
@@ -26,6 +26,7 @@ function GoogleSignIn(props: Props) {
             props.onAfterLogin();
         }).catch(() => {
             googleId = null;
+            localStorage.removeItem("googleId");
         })
     };
 
--- a/src/components/GoogleSignIn/GoogleSignIn.tsx
+++ b/src/components/GoogleSignIn/GoogleSignIn.tsx
@@ -24,6 +24,8 @@ function GoogleSignIn(props: Props) {
         setGoogleId(response.tokenId);
         api.setGoogle(response.tokenId).then(() => {
             props.onAfterLogin();
+        }).catch(() => {
+            googleId = null;
         })
     };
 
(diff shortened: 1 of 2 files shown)


Ekwav:
this doesn't fix #71 
I think, it has to at least delete the token from localstorage as well

src/components/GoogleSignIn/GoogleSignIn.tsx
```diff
@@ -24,6 +24,8 @@ function GoogleSignIn(props: Props) {
         setGoogleId(response.tokenId);
         api.setGoogle(response.tokenId).then(() => {
             props.onAfterLogin();
+        }).catch(() => {
+            googleId = null;
```
Ekwav, line 28: this doesn't fix #71
I think, it has to at least delete the token from localstorage as well

Ekwav (changes requested):

src/components/GoogleSignIn/GoogleSignIn.tsx
```diff
@@ -24,6 +24,9 @@ function GoogleSignIn(props: Props) {
...
+            localStorage.removeItem("googleId");
```
Ekwav, line 29: And rerender the page

Ekwav (changes requested):

Ekwav:
You caneasily test yourself by setting "googleId" in localstorage  to some random value. (thus invalid and login should be retriggered)

src/components/GoogleSignIn/GoogleSignIn.tsx
matthias-luger: The GoogleLogin-Component does rerender if the googleId is set to null.
If you refer to the parent-Component. I cant, dont want and dont need to rerender this.
The onAfterLogin-Function is never executed, so the parent knows, it hasnt logged in

PR #78, opened 2021-04-12
copy buttons

### LINKED ISSUE #57: Copy Buttons
Add buttons to copy an viewauction command (/viewauction uuid) and player auctions (/ah playername)

matthias-luger:
fixes #57

### DIFF
--- a/src/api/ApiHelper.tsx
+++ b/src/api/ApiHelper.tsx
@@ -405,7 +405,6 @@ function initAPI(): API {
                     resolve();
                 },
                 reject: (error: any) => {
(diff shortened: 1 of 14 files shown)


Ekwav:
The issues asks for a copy button on the auctionDetails page like it is on the playerDetails page.
Please don't put it inline with other auctionDetails it is very hard to find that way. (I searched almost a minute, and knew it exists, others might not :smile: )

Ekwav:
The copy button on playerDetails has to copy the username not the users uuid.

PR #96, opened 2021-05-15
flickering reduced

### LINKED ISSUE #92: Reduce flicker
All elements should be created with fixed height (and width).
They may be filled with placeholders and later replaced with text.
Placeholders are just gray rectangles.
This reduces irritating flickering when loading and gives the perception that the page is loading faster

matthias-luger:
fixes #92


Ekwav:
The Cumulative Layout Shift on /players/ is still 0,36 (way to high)

Ekwav (approved):

PR #140, opened 2021-06-03
fixed names in recent auctions

matthias-luger:


Ekwav (approved):
react is weird, nothing new

PR #150, opened 2021-06-07
changed endless scrolling to stop at 0

### LINKED ISSUE #122: Add request to endless scroll
The backend is about to be broken up into multiple smaller ones.
As a result it is possible that a request returns fewer than 20 (or now 10) auctions while there are still more.
As long as the returned auction count is not 0, the request (with higher offset) should be repeated.

matthias-luger:
fixes #122


Ekwav (approved):

Ekwav (approved):

Ekwav:
Still riddling why that are two components

PR #153, opened 2021-06-07
implemented properties

### LINKED ISSUE #152: Move settings to special file
Important settings like the `server-url` should be moved to a static place like a single `.json` or into the `index.hmtl`.
This is necessary for me to be able to programmaticly change them for testing before and during updates. (running with different domains)

matthias-luger:
fixes #152


src/utils/PropertiesUtils.tsx
```diff
@@ -0,0 +1,5 @@
+import properties from '../properties.json';
```
Ekwav, line 1: by importing the `.json` file it gets bundled into `/static/js/main.f9371155.chunk.js` this was the case previously already.
Please allow the object (JSON) to be in `index.html` this way an extra roundtrip to the backend is prevented.

Ekwav (changes requested):

Ekwav (approved):

PR #154, opened 2021-06-08
catched error if cache is used in private window

### LINKED ISSUE #148: Support Firefox
The page should work in firefox as 10% of our users are using Firefox.

matthias-luger:
fixes #148 

(last fix didnt work)


src/utils/CacheUtils.tsx
```diff
@@ -16,9 +16,10 @@ let cacheUtils: CacheUtils = {
                         }
                     }
                     resolve(null);
+                }).catch(() => {
+                    resolve(null);
```
Ekwav, line 20: the `.catch(` has to be used in `checkForCacheClear` as well

Ekwav (changes requested):

Ekwav (approved):

PR #155, opened 2021-06-08
error on load

### LINKED ISSUE #151: Stop loading if error occured
The frontend should either stop displaying loading or retry the request if an error occurred. 
This is mainly targeted at invalid/empty responses.

matthias-luger:
fixes #151 

Where do you really see the problem?
Pretty much everywhere the user gets notified if an error occurs. @Ekwav

### DIFF
--- a/src/api/HttpHelper.tsx
+++ b/src/api/HttpHelper.tsx
@@ -37,12 +37,26 @@ function sendRequest(request: ApiRequest): Promise<void> {
             .then(response => {
                 if (response.ok && response.body === null) {
                     request.reject();
+                    return;
                 }
                 if (!response.ok) {
                     request.reject();
+                    return;
                 }
-                return response.json();
+
+                let parsed;
+                try {
+                    parsed = response.json();
+                } catch (error) {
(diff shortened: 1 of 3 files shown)


Ekwav:
The issue occurs on invalid responses (like status code 200 but empty)
I am trying to find the issue why responses are empty but were unsuccessful so far.
The frontend should treat it like any other `error` response

src/api/HttpHelper.tsx
```diff
@@ -34,8 +34,15 @@ function sendRequest(request: ApiRequest): Promise<void> {
...
-            .then(response => response.json())
-            .then(parsedResponse => {
+            .then(response => {
+                if (response.ok && response.body === null) {
+                    request.reject();
+                }
+                if (!response.ok) {
+                    request.reject();
+                }
+                return response.json();
```
Ekwav, line 44: an exception from failing `JSON.parse` is not catched

Ekwav (changes requested):

Ekwav (approved):

PR #168, opened 2021-06-15
Revert "fix"

Ekwav:
Reverts matthias-luger/hypixel-react#167
breaks the item name


Ekwav:
Also breaks scrolling

PR #176, opened 2021-06-16
both filter now work

matthias-luger:


Ekwav (changes requested):
please add a cypress test to make sure this doesn't occur again.

matthias-luger:
> please add a cypress test to make sure this doesn't occur again.

i beg you not

src/components/BidList/BidList.tsx
```diff
@@ -90,6 +90,10 @@ function BidList(props: Props) {
             })
             setBids(bids);
 
+            if(bids.length < 10 && newBids.length !== 0){
```
Ekwav, line 93: This change is redundant, please share code with AuctionList.tsx

Ekwav (changes requested):

src/components/BidList/BidList.tsx
matthias-luger: historisch bedingt

Ekwav (approved):
looks ok

PR #179, opened 2021-06-19
fixed httphelper

### LINKED ISSUE #171: Auction details missing player names
Most of the time player names are missing from the auction details page.
I will add the names to the bids array as a new object that will also include a preview for the player head. (A smaller version of the head base64 encoded)

matthias-luger:
fixes #171


src/api/HttpHelper.tsx
```diff
@@ -1,19 +1,20 @@
...
 
 function sendRequest(request: ApiRequest): Promise<void> {
+    request.mId = getNextMessageId();
     let requestString = JSON.stringify(request.data);
     let headers = { 'ConId': getOrGenerateUUid() };
-    var url = `${commandEndpoint}/${request.type}/${Base64.encode(requestString)}`;
-    if (request.mId)
-        url += `/${request.mId}`;
+    let url = `${commandEndpoint}/${request.type}/${Base64.encode(requestString)}`;
+    url += `/${request.mId}`;
```
Ekwav, line 17: This makes the cloudflare cache absolutely useless

Ekwav (changes requested):

src/api/HttpHelper.tsx
matthias-luger: no cache problem any more? Best fix ever!

src/api/HttpHelper.tsx
```diff
@@ -1,19 +1,19 @@
...
 const commandEndpoint = getProperty("commandEndpoint");
 let requests: ApiRequest[] = [];
 
 function sendRequest(request: ApiRequest): Promise<void> {
+    request.mId = getNextMessageId();
     let requestString = JSON.stringify(request.data);
     let headers = { 'ConId': getOrGenerateUUid() };
     var url = `${commandEndpoint}/${request.type}/${Base64.encode(requestString)}`;
-    if (request.mId)
-        url += `/${request.mId}`;
```
Ekwav, line 16: the cloudflare cache has to be skipped if the command is valid for less than half an hour.
`sendRequestLimitCache` used the mId to for this but now can't

src/api/HttpHelper.tsx
```diff
@@ -99,7 +101,7 @@ function getOrGenerateUUid(): string {
 }
 function findForEqualSentRequest(request: ApiRequest) {
     return requests.filter(r => {
-        return r.type === request.type && r.data === request.data
+        return r.type === request.type && r.data === request.data && r.mId !== request.mId;
```
Ekwav, line 104: what is this check? It will lead to resending of exactly matching `mId`s.

Ekwav (changes requested):

src/api/HttpHelper.tsx
matthias-luger: The method should find "equal" Requests. Not the exact request it got as a parameter. This parameter is handled before.

Ekwav (approved):

PR #181, opened 2021-06-19
implemented flip based auction lookup in flipper

### LINKED ISSUE #174: Display flipBased command
The command `flipBased` allows you to retrieve the auctions that were used to calculate the median price of a flip.
The list should be available in the flipper UI over a special button.

matthias-luger:
fixes #174


Ekwav (approved):
Amazing :D

PR #182, opened 2021-06-19
auction and bid list now have 2 columns

### LINKED ISSUE #170: Make auction and bid list 2 columns
There is more than enough room to display the auction and bid list in two columns on screen sizes larger than 1200px.
Also please reuse parts of the code in both components.

matthias-luger:
fixes #170 

the reused code is bad but i would like to refactor the lists in a seperate ticket please


Ekwav (approved):
It is possible that the right column is mistaken for the players bids.

PR #184, opened 2021-06-19
implemented max cost filter

### LINKED ISSUE #164: Add max cost to flipper
To allow lower net worth users to use the flipper there should be an additional filter `max cost`.
To organize the filters better, put them in two columns on screens larger than 800px.

matthias-luger:
fixes #164

with the checkbox formfield 2 columns look kinda weird in my opinion
didnt implement just yet


src/components/Flipper/Flipper.tsx
```diff
@@ -130,6 +130,9 @@ function Flipper() {
                 if (flipperFilter?.minProfit && flipperFilter.minProfit >= (auction.median - auction.cost)) {
                     return false;
                 }
+                if (flipperFilter?.maxCost && flipperFilter.maxCost > auction.cost) {
```
Ekwav, line 133: This is the worng way around, you implemented `minCost`

Ekwav (changes requested):
Maybe the two input fields should go in one line and the check boxes in the next line.
I think that would look better

Ekwav (approved):

PR #185, opened 2021-06-19
use name from itemDetails

### LINKED ISSUE #163: Display correct Item Name
The `/item/` site currently converts the tag to `guess` the item name.
This is often wrong. The first entry in the `names` array of `itemDetails` should be used instead. 
The names are sorted in a way that should bring the correct name to the front.
Example: https://sky.coflnet.com/i/jerry-chine

matthias-luger:
fixes #163


Ekwav (changes requested):
This doesn't solve the issue.
http://localhost:3000/item/JERRY_STAFF
Should display `Jerry-chine Gun` but doesn't

Ekwav (approved):

Ekwav (approved):

PR #232, opened 2021-07-09
implemented swipe

### LINKED ISSUE #221: Swipe between auctions and bids
Enable swipe gesture on `/player` to switch between auction and bids

matthias-luger:
fixes #221


Ekwav (approved):
A smooth transition would be cooler but whatever.

PR #234, opened 2021-07-09
use img prop in fullsearch

### LINKED ISSUE #223: Display prefetched images in search
`fullSearch` returns a new property `img` wich contains a `base64 encoded` png  or `null`.
In case there is an image available it should be displayed immediately an not request should be made.

matthias-luger:
fixes #223


Ekwav (approved):
branch has conflicts with master

PR #240, opened 2021-07-10
navbar-rework

### LINKED ISSUE #161: Always display menu items on large screen
There is a lot of empty space when viewing the site on full hd and above.
This space should be used to always display the menu items.
This reduces friction and gives new sites like the flipper more visibility.

matthias-luger:
fixes #161


Ekwav (changes requested):
This has two very severe bugs
* the side bar overlaps with the content (there should be a fixed padding on larger screens)
* the hamburger icon vanishes sometimes (found this when changing the screen size)
image
I suggest that the hamburger icon is always in the upper left and just overlayed by the menu

matthias-luger:
@Ekwav The icon "vanishes" at 1400px width. Thats the time where i always show a collapsed navbar. There isnt really a reason for the hamburger icon as you can hover over the navbar and it kinda looked off. Where should be the hamburger icon in your opinon (width > 1400px)

Ekwav:
That's the whole point I can't hover over the menu because it isn't visible.
The screenshot snipped shows Starts at the left of the screen there is no menu nor hamburger icon.
It should be impossible to not being able to reach the menu.
My suggestion is to always have the hamburger icon in the upper right to work around a failing menu.

Ekwav (approved):
Fantastic :)

PR #252, opened 2021-07-13
search enchantment

matthias-luger:
fixes #nothing


Ekwav (commented):
Does not work when already viewing enchanted books

PR #272, opened 2021-07-17
premium feature out of the window 2 (second worst merge)

### LINKED ISSUE #262: Allow non premium users to use all filters
I still think that non-premium users should have a chance to use the `BIN` and `not SOLD` filters. 
Please allow them to check the filter after 1-5 min on the site (preferably with a countdown).
It should not be too annoying to use the free flipper.
And after more than a minute everyone got it, that premium would be better but decided not to get it.
This is the same method as hypixel uses for daily rewards where you have to watch a 30sec youtube video before you can claim the reward.

matthias-luger:
fixes #262

### DIFF
--- a/src/components/Flipper/FlipperFilter/FlipperFilter.tsx
+++ b/src/components/Flipper/FlipperFilter/FlipperFilter.tsx
@@ -11,6 +11,7 @@ interface Props {
 }
 
 let FREE_PREMIUM_FILTER_TIME = new Date(new Date().getTime() + (1000 * 60 * 5));
+let FREE_LOGIN_FILTER_TIME = new Date(new Date().getTime() + (1000 * 60 * 6));
 
 function FlipperFilter(props: Props) {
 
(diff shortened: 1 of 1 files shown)


src/components/Flipper/FlipperFilter/FlipperFilter.tsx
```diff
@@ -59,52 +63,62 @@ function FlipperFilter(props: Props) {
...
-        let filter = getCurrentFilter();
-        filter.onlyUnsold = isActive;
-        props.onChange(filter);
-    }
-
     return (
         <div>
             <Form style={{ marginBottom: "5px" }} >
-                {!props.isLoggedIn ?
+                {!props.isPremium && !freePremiumFilters ?
```
Ekwav, line 109: This should still require a login and/or display that login is an alternative to unlock it

Ekwav (changes requested):

Ekwav (approved):

PR #275, opened 2021-07-17
rework

### LINKED ISSUE #242: Improve flipper performance
The flipper gets slow (and sometimes crashes the tab) after more than a thousand flips.
This should be avoided at all cost.
The main bottleneck seems to be cpu time.
This is likely caused by aggregate functions such as `filter` and `map` which always process the whole list.
Please improve the flippers performance to support at least 10k flips, and start deleting flips over that mark.

matthias-luger:
fixes #242


Ekwav (approved):
I guess this is more performant

PR #276, opened 2021-07-17
auction details in flipper

### LINKED ISSUE #160: Display auction details inside the flipper
As the flipper is usually used on large PC screens it makes sense to not open the auction details in a new window but rather display them within the flipper below the scrollist.
This is especially important for auctions with NBT data.
There may be an additional button to open it in a new window.

matthias-luger:
fixes #160


Ekwav (changes requested):
* There should be a retry if the auction wasn't yet found. (after 2-5 seconds) 
* Clicking on things in auction details may also open a new tab

src/components/Flipper/Flipper.tsx
```diff
@@ -215,7 +204,7 @@ function Flipper() {
                         {!isLoggedIn ?
                             <div>
                                 <h2>Free auction house flipper preview - hypixel skyblock ah history</h2>
-                                You need to be logged and have Premium to have all features unlocked.<br/><br/>
+                                You need to be logged and have Premium to have all features unlocked.
```
Ekwav, line 207: this looks like not on purpose (was recently changed in another PR)

Ekwav (changes requested):

PR #281, opened 2021-07-17
fix 2

### LINKED ISSUE #282: The countdown of the flipper filter doesnt work

matthias-luger:
fixes #282


Ekwav (approved):
still think the delay is to long

PR #297, opened 2021-07-23
added lowestBin and sellerName

### LINKED ISSUE #288: Display lowest bin and seller in the flipper
Flips got extended to include the seller name and the lowest bin price for a given item.
Both should be displayed withing the flipper.
When the lowest price is clicked `/item/{itemTag}/active` (or whatever the active endpoint will end up being located at) should be opened with the filters `Bin` and `sortOrder` `lowestPrice` should be opened. 
Depends on  #227

matthias-luger:
fixes #288


Ekwav (approved):
lgtm

PR #300, opened 2021-07-24
fix flipper filter timeouts

### LINKED ISSUE #291: Countdown on /flipper is out of sync
The countdown on /flipper starts when the flipper JavaScript is loaded. However the filter features are only enabled once the /flipper was open for 5 min.
The time displayed should not differ from when the checkbox is enabled by more than a second.

matthias-luger:
fixes #291

### DIFF
--- a/src/components/Flipper/FlipperFilter/FlipperFilter.tsx
+++ b/src/components/Flipper/FlipperFilter/FlipperFilter.tsx
@@ -11,7 +11,7 @@ interface Props {
     isPremium?: boolean
 }
 
-let FREE_PREMIUM_SPAN = 1000 * 60 * 0.1;
+let FREE_PREMIUM_SPAN = 1000 * 60 * 5;
 let FREE_LOGIN_SPAN = 1000 * 60 * 6;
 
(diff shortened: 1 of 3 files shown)


src/components/Flipper/FlipperFilter/FlipperFilter.tsx
```diff
@@ -3,18 +3,19 @@ import { Form } from 'react-bootstrap';
...
+import { v4 as generateUUID } from 'uuid';
 
 interface Props {
     onChange(filter: FlipperFilter),
     isLoggedIn?: boolean,
     isPremium?: boolean
 }
 
-let FREE_PREMIUM_SPAN = 1000 * 60 * 5;
+let FREE_PREMIUM_SPAN = 1000 * 60 * 0.1;
```
Ekwav, line 14: I think you made this by mistake and would yell at me when I merge it

Ekwav (approved):

src/components/Flipper/FlipperFilter/FlipperFilter.tsx
Flou21, line 14: const

PR #308, opened 2021-07-27
track users who have already given consent before

### LINKED ISSUE #309: old users are not tracked
Old users who have already given consent are not tracked anymore

matthias-luger:
fixes #309


Ekwav (approved):
Following GDPR we also need a way to allow users to opt out of tracking asap.

PR #313, opened 2021-07-27
add dockerfile only frontend for temp env

Flou21:
- added dockerfile, removed unused stuff
- budockerfile only builds frontend


Flou21:
created temp environment: https://temp-env-sky-313.flou21.de

PR #315, opened 2021-07-27
add dockerfile to support temp env

Flou21:
- added dockerfile, removed unused stuff
- budockerfile only builds frontend
- check if temp env, then call preview system api
- removed protocol from hostname check


Flou21:
created temp environment: https://temp-env-sky-315.flou21.de

Ekwav (approved):
Tests don't run against the temp domain

PR #329, opened 2021-07-29
implement number test

matthias-luger:


Ekwav (commented):
What does this change?

PR #335, opened 2021-07-30
removed cookie libary

matthias-luger:
removed cookie libary 
used the better js-cookie libary


Flou21:
created temp environment: https://temp-env-sky-335.flou21.de

Ekwav (approved):

PR #341, opened 2021-07-31
fix apiinfo page

matthias-luger:


Flou21:
created temp environment: https://temp-env-sky-341.flou21.de

PR #378, opened 2021-08-10
add production.spec.ts test

Flou21:
started with tests for production, only one but more than 0


Ekwav (approved):
LGTM!
but has merge conflicts

PR #402, opened 2021-08-14
only use staging on port 3000

### LINKED ISSUE #400: only use staging system when served from port 3000
Currently the staging back-end is used when the frontend detects it is running on localhost.
This should be changed to only use staging when running on localhost:3000 and otherwise use the current host ( /command)

matthias-luger:
fixes #400


Flou21:
[CI] created a pr environment: https://pr-env-sky-402.flou21.de

Ekwav (approved):

PR #406, opened 2021-08-14
fix lowest bin click

### LINKED ISSUE #404: Fix second lowest bin click
The second lowest bin click has to be clicked twice to open the item page.
It should behave the same as the lowest bin itself and open it right away.
I guess there is a missing preventDefault

matthias-luger:
fixes #404


Ekwav (approved):

Flou21:
[CI] created a pr environment: https://pr-env-sky-406.flou21.de

PR #455, opened 2021-09-05
implement white/blacklist of the backend

### LINKED ISSUE #432: Implement full filter support for the blacklist feature
Users want to blacklist specific items e.g. uncommon enchanted books. For that, the blacklist feature should support the filter feature.

matthias-luger:
fixes #432

### DIFF
--- a/src/api/ApiHelper.tsx
+++ b/src/api/ApiHelper.tsx
@@ -513,7 +513,7 @@ function initAPI(): API {
             minVolume: filter.minVolume || 0,
             maxCost: filter.maxCost || 0,
             filter: {
-                BIN: filter.onlyBin || false
+                BIN: filter.onlyBin ? filter.onlyBin.toString() : "false"
             }
         }
 
--- a/src/api/ApiHelper.tsx
+++ b/src/api/ApiHelper.tsx
@@ -8,6 +8,7 @@ import { enchantmentAndReforgeCompare } from "../utils/Formatter";
 import { googlePlayPackageName } from '../utils/GoogleUtils'
 import { toast } from 'react-toastify';
 import cacheUtils from "../utils/CacheUtils";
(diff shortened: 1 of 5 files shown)


matthias-luger:
@Ekwav I think the UI works and sends the correct data but the new backend doesnt work.

Ekwav:
There was indeed a bug when specifying a blacklist with an empty array. 
I fixed it.  
This PR should target `develop` it would break production

src/properties.js
```diff
@@ -1,7 +1,7 @@
 let properties = {
-    "commandEndpoint": window.location.host === 'localhost:3000' || window.location.hostname.includes("pr-env-sky-") ? "https://skyblock-backend.coflnet.com/command" : "/command",
-    "apiEndpoint": window.location.host === 'localhost:3000' || window.location.hostname.includes("pr-env-sky-") ? "https://skyblock-backend.coflnet.com/api" : "/api",
-    "websocketEndpoint": "wss://skyblock-backend.coflnet.com/skyblock",
+    "commandEndpoint": window.location.host === 'localhost:3000' || window.location.hostname.includes("pr-env-sky-") ? "https://sky-commands.coflnet.com/command" : "/command",
+    "apiEndpoint": window.location.host === 'localhost:3000' || window.location.hostname.includes("pr-env-sky-") ? "https://sky-commands.coflnet.com/api" : "/api",
+    "websocketEndpoint": "wss://sky-commands.coflnet.com/skyblock",
```
Ekwav, line 4: This PR should target `develop` it would break production

Ekwav (changes requested):

Ekwav (changes requested):
Some issues I found:
- When any filter is changed all flips are deleted.
- a date filter is displayed as timestamp
- the filters are only for "aspect of the end" it should request "*"

PR #456, opened 2021-09-05
implement correct name for based auction

### LINKED ISSUE #441: Display correct item name for reference auctions
Currently the title of the new item is used for all of them which is wrong

matthias-luger:
fixes #441

### DIFF
--- a/src/utils/Parser/APIResponseParser.tsx
+++ b/src/utils/Parser/APIResponseParser.tsx
@@ -31,7 +31,7 @@ export function parseAuction(auction: any): Auction {
         end: new Date(auction.end),
         item: {
             tag: auction.tag,
-            name: auction.itemName || auction.ItemName
+            name: auction.itemName || auction.name
         },
         startingBid: auction.startingBid,
         highestBid: auction.highestBid,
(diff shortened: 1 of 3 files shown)


matthias-luger:
It would be cool if the property could be renamed from ItemName to itemName like it is everywhere else.

Ekwav:
It is not called `ItemName`, it is called `name` should I rename it to `itemName` ?

Ekwav:
also this MR targets should target `dev` because the property is only available in commands

Ekwav:
also this MR targets should target `dev` because the property is only available in commands

src/utils/Parser/APIResponseParser.tsx
```diff
@@ -31,7 +31,7 @@ export function parseAuction(auction: any): Auction {
         end: new Date(auction.end),
         item: {
             tag: auction.tag,
-            name: auction.itemName
+            name: auction.itemName || auction.ItemName
```
Ekwav, line 34: > It is not called `ItemName`, it is called `name` should I rename it to `itemName` ?
https://github.com/Coflnet/hypixel-react/pull/456#issuecomment-917605441

Ekwav (changes requested):

Ekwav (approved):

PR #458, opened 2021-09-05
Authorize mod page

### LINKED ISSUE #453: Create /authCon
To authenticate the mod a new page is required that receives the connectionId is required.  
It should 
1. send the "conId" query parameter via the "authCon" command  (as number)
2. display the information that the Connection is now authorized.

matthias-luger:
fixes #453

### DIFF
--- a/src/api/ApiHelper.tsx
+++ b/src/api/ApiHelper.tsx
@@ -768,7 +768,7 @@ function initAPI(): API {
     let authenticateModConnection = (conId: number) : Promise<void> => {
         return new Promise((resolve, reject) => {
 
-            httpApi.sendApiRequest({
+            websocketHelper.sendRequest({
                 type: RequestType.AUTHENTICATE_MOD_CONNECTION,
                 data: conId,
                 resolve: function () {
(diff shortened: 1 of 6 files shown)


src/routes.tsx
```diff
@@ -54,6 +55,7 @@ setTimeout(() => {
   Success.preload();
   Ref.preload();
   ApiInfo.preload();
+  AuthMod.preload();
```
Ekwav, line 58: the authMod page can't be reached without a direct link, thus it makes no sense to preload it

src/api/ApiHelper.tsx
```diff
@@ -765,6 +765,23 @@ function initAPI(): API {
         })
     }
 
+    let authenticateModConnection = (conId: number) : Promise<void> => {
+        return new Promise((resolve, reject) => {
+
+            httpApi.sendApiRequest({
```
Ekwav, line 771: command needs the websocket session to know which user to authenticate

Ekwav (changes requested):

PR #469, opened 2021-09-11
save filter on item change

### LINKED ISSUE #465: Save valid filters on item change
Keep valid filters after you change the item

matthias-luger:
fixes #465


Ekwav (approved):
LGTM

PR #492, opened 2021-09-25
Reduce subflip commands

### LINKED ISSUE #489: only send subFlip once or when something changed
The `subFlip` is currently sent about 5 times on page load.
This should be reduced to once.

matthias-luger:
fixes #489


Ekwav (approved):
No clue why but this works

PR #512, opened 2021-10-08
Fix bugged payment backend

### LINKED ISSUE #510: Fix premium purchase
Requires #509 
Premium can be bought via skyblock-backend.coflnet.com all payment related commands should be sent to that websocket.

matthias-luger:
fixes #510 
Fixes the payment bugs that were introduced on purpose because a not working the backend had to be set live for whatever reason...

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15370,13 +15370,11 @@
             }
         },
         "react-toastify": {
-            "version": "6.2.0",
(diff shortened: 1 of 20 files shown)


Ekwav:
The reasons were, instability, bugs and missing features of the other system.  
We discussed the release multiple times.   
PS: You target the wrong branch.

PR #524, opened 2021-10-10
use utc in time parsing

### LINKED ISSUE #521: The end time on auctionDetails is wrong
The time on `/acution/{uuid}` doesn't parse the end time as `UTC` its using local time, which is wrong.
It should read the time as UTC and then convert it to the local timezone

matthias-luger:
fixes #521


Ekwav (approved):
"It works"

PR #574, opened 2021-11-02
removed defaultValue from the filters

### LINKED ISSUE #568: better filter searching for ah viewer
make the filters be empty when you first select it, and remove the text imput for enchant levels, hpb, recomb, bin, stars.

matthias-luger:
reimplemented select without search
fixes #568

### DIFF
--- a/src/components/FilterElement/FilterElement.tsx
+++ b/src/components/FilterElement/FilterElement.tsx
@@ -13,13 +13,14 @@ import api from '../../api/ApiHelper';
 
 // has to be redefined because global types from react-app-env are not accessable
 export enum FilterType {
-    Equal = 1,
+    EQUAL = 1,
     HIGHER = 2,
     LOWER = 4,
     DATE = 8,
     NUMERICAL = 16,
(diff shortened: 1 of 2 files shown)


src/components/FilterElement/FilterElement.tsx
```diff
@@ -162,15 +169,21 @@ function FilterElement(props: Props) {
...
+                                        </Form.Control>
+                                        : hasFlag(props.options.type, FilterType.EQUAL)
+                                            ? <Typeahead
+                                                id={props.options.name}
+                                                style={{ display: "block" }}
+                                                className="select-filter"
+                                                onChange={selected => updateSelectFilter(selected)}
+                                                options={props.options?.options}
+                                                labelKey={convertTagToName}
+                                                autoselect={false}
```
Ekwav, line 184: What does this setting do?

matthias-luger: With autoselect={true} the first option is put into the select field and has to be deleted by the user if he wants to select something else.

PR #575, opened 2021-11-03
first attempt to fix pages that broke because of chunk loading errors

### LINKED ISSUE #543: Detect and mitigate broken page
Whenever there is an update of the frontend JS chunks are added and others are removed.
The website then sometimes enters a broken state.
An attempt should be made to mitigate this.
Some possible solutions I came up with are 
1. Clearing cache and Force reloading 
2. Loading the page with ?t=XY appended to the url to bypass the cloudflare cache.

Either of this option may or may not be triggered automatically or only after the user pressing a button.

matthias-luger:
fixes #543

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -5,11 +5,14 @@
     <script async src="/preScript.js"></script>
     <script async src="/MinecraftColorCodes.3.7.js"></script>
     <script>
-        // check if react was correctly loaded after 5 seconds
-        setTimeout(() => {
-            debugger;
-            let div = document.getElementById('react-container');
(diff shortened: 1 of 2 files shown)


matthias-luger:
This is my first attempt to fix the problem. 
As I can't reliably reproduce it, I couldn't test it tough.

Ekwav (commented):
Did you test this with simulated very slow network?

Ekwav (approved):
I guess this is better. Can't find an obvious error, let's put it to the test.

PR #577, opened 2021-11-05
Implemented the low suppy page

matthias-luger:
Added low supply page to navigation
Reworked the getItemImageUrl function


matthias-luger:
@Ekwav Under which URL ist the indexer available?

Ekwav:
> @Ekwav Under which URL ist the indexer available?

It's only available through commands (and this endpoint isn't yet proxied)

PR #588, opened 2021-11-13
extracted filtertype into own file

### LINKED ISSUE #587: Default value of the search select
The default value of the search select fields doesnt seem to work anymore. 
Exampe: https://sky.coflnet.com/item/ENCHANTED_BOOK?itemFilter=eyJFbmNoYW50bWVudCI6InByb3RlY3Rpb24iLCJFbmNoYW50THZsIjoiNiJ9&apply=true&range=week

matthias-luger:
fixes #587


src/components/ItemFilter/ItemFilter.tsx
```diff
@@ -219,11 +220,19 @@ function ItemFilter(props: Props) {
         onFilterChange(newFilter);
     }
 
+    function hasFlag(full?: FilterType, flag?: FilterType) {
```
Ekwav: Isn't this method now duplicated?
Could it be shared between components?

Ekwav (changes requested):

PR #598, opened 2021-11-20
fixed bug with the tag property of the blacklisted items

### LINKED ISSUE #593: Blacklist Tag entries are null
I got bug reports from users with `null` on their blacklist multiple times (and empty filters).  
I suspect the item tag gets lost somewhere.

matthias-luger:
fixes #593


src/components/MainApp/MainApp.tsx
```diff
@@ -44,6 +45,8 @@ export function MainApp(props: any) {
                 localStorage.setItem("lastReloadTime", new Date().getTime().toString());
             }
         }
+
+        startMigrations();
```
Ekwav, line 49: I think this should not trigger automatically or at least tell the user in a way that he should double check the new blacklist

Ekwav (changes requested):

PR #599, opened 2021-11-20
may fix the issue?

### LINKED ISSUE #595: A lot of JWT expired errors
image
Besides the fact that I am on a page that does not require a login, that error should only appear once

matthias-luger:
maybe fixes #595


Ekwav (approved):
I have no idea what it does but let's try

PR #607, opened 2021-11-27
implemented finders

### LINKED ISSUE #604: Add option to select finders
There are now more flip finder options. 
They are identified by 
```
            FLIPPER = 1,
            SNIPER = 2,
            SNIPER_MEDIAN = 4
```
the key in settings is called `finders` and every flip has the property `finder`

Description for each option:
Flipper is the classical flip finding algorithm using the `Skyblock AH history` database. It searches the history for similar items but searching for references takes time thus this is relatively slow.
Sniper is a classical sniping algorithm that stores prices in a dictionary grouped by any relevant modifiers. It only outputs flips that are below lbin and median for a combination of relevant modifiers. Its faster by about 3000x but may not find as many flips as the flipper.
Sniper median uses the same algorithm as Sniper but doesn't require the item to be below lowest bin and only 10% below the median sell value.

matthias-luger:
fixes #604

### DIFF
--- a/src/api/ApiHelper.tsx
+++ b/src/api/ApiHelper.tsx
@@ -534,13 +534,16 @@ function initAPI(): API {
                 volume: !flipSettings.hideVolume,
                 extraFields: flipSettings.maxExtraInfoFields,
                 profitPercent: !flipSettings.hideProfitPercent
-            }
+            },
+            finders: flipSettings.finders?.reduce((a, b) => +a + +b, 0)
         }
 
         if (filter.onlyBin) {
             requestData.filters = { Bin: "true" };
         }
 
+        console.log(requestData);
+        
         websocketHelper.subscribe({
             type: RequestType.SUBSCRIBE_FLIPS,
             data: requestData,
--- a/src/api/ApiHelper.tsx
+++ b/src/api/ApiHelper.tsx
@@ -541,8 +541,6 @@ function initAPI(): API {
         if (filter.onlyBin) {
             requestData.filters = { Bin: "true" };
         }
-
-        console.log(requestData);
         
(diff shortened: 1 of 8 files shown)


src/api/ApiHelper.tsx
```diff
@@ -534,13 +534,16 @@ function initAPI(): API {
...
-            }
+            },
+            finders: flipSettings.finders?.reduce((a, b) => +a + +b, 0)
         }
 
         if (filter.onlyBin) {
             requestData.filters = { Bin: "true" };
         }
 
+        console.log(requestData);
```
Ekwav: please remove

src/components/Flipper/FlipCustomize/FlipCustomize.tsx
```diff
@@ -191,10 +207,22 @@ function FlipCustomize() {
...
+        return FLIP_FINDERS.filter(option => settings.finders?.some(finder => finder.toString() === option.value))
+    }
+
     let useLowestBinHelpElement = (
         <p>By enabling this setting, the lowest BIN is used as the estimated selling price to calculate your profit. That can lead to profitable flips being estimated way too low (even as a loss). We recommend using the median to calculate the profit.</p>
     );
 
+    let flipFindersElement = (
+        <ul>
+            <li><b>Flipper </b>is the classical flip finding algorithm using the Skyblock AH history database. It searches the history for similar items but searching for references takes time thus this is relatively slow.</li>
```
Ekwav: could these tooltips be displayed on the actual element itself? 
image
It will probably grow in the not to distant future and is already very hard to read

Ekwav (changes requested):

PR #608, opened 2021-11-29
Implemented updateSettings websocket listening

### LINKED ISSUE #592: accept the `settingsUpdate` command
Each frontend tab overwrites each other tabs-settings, also the mod will likely get some settings changing ability that should sync with the website.
Therefor the website should accept and save settings changes from the backend.
the `settingsUpdate` has the same format as `subFlip` and is sent whenever the settings are changed.

matthias-luger:
fixes #592

### DIFF
--- a/src/api/ApiHelper.tsx
+++ b/src/api/ApiHelper.tsx
@@ -12,10 +12,9 @@ import { checkForExpiredPremium } from "../utils/ExpiredPremiumReminderUtils";
 import { getFlipCustomizeSettings } from "../utils/FlipUtils";
 import { getProperty } from "../utils/PropertiesUtils";
 import { Base64 } from "js-base64";
-import { FLIPPER_FILTER_KEY, FLIP_CUSTOMIZING_KEY, RESTRICTIONS_SETTINGS_KEY, setSetting, setSettingsChangedData } from "../utils/SettingsUtils";
+import { setSettingsChangedData } from "../utils/SettingsUtils";
 
 let connectionId = null;
-let changer = null;
 
 function initAPI(): API {
 
@@ -531,7 +530,7 @@ function initAPI(): API {
             visibility: {
                 cost: !flipSettings.hideCost,
                 estProfit: !flipSettings.hideEstimatedProfit,
-                lBin: !flipSettings.hideLowestBin,
+                lbin: !flipSettings.hideLowestBin,
                 slbin: !flipSettings.hideSecondLowestBin,
                 medPrice: !flipSettings.hideMedianPrice,
                 seller: !flipSettings.hideSeller,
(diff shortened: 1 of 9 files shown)


matthias-luger:
Should work for FlipCustomization and FlipperFilters.
Doesnt work for FlipRestrictions yet (blacklist/whitelist). I send you the tag and filter for the blacklist/whitelist, because thats all you need, but the UI needs the iconUrl and the itemName as well. I could fetch the name for every settingsUpdate you send me (from another changer) or I send you the name in the subFlip message. 

What method would you prefer? @Ekwav

Ekwav:
> FlipRestrictions

figure out whats new and add it  to the old black/whitelist and only load the new required information

Ekwav (changes requested):
Settings are constantly overwritten (reset ) by other tabs
Updated blacklist isn't used (never displayed)
merge conflicts

matthias-luger:
@Ekwav 
One tab sends a perfectly fine subFlip-Command and the other gets a empty settingsChanged. This breaks things...
Other than that I couldnt replicate your issues. Please make sure to clear every cache.
Tab A:
image

Tab B.
image

Ekwav (changes requested):
The website sometimes doesn't accept changes for a yet unkown reason.
Could you add a temporary option to reset the settings to the one before the update (copy the local storage entry to a separate key if it doesn't exist already and add an option to restore them)

matthias-luger:
@Ekwav I tried to reproduce your problems but it seems like settingsUpdate is not sent any more at all? Did I mess something up?

Ekwav:
> @Ekwav I tried to reproduce your problems but it seems like settingsUpdate is not sent any more at all? Did I mess something up?

Whupsi, that broke with the recent attempts to stabilize the page. I added it back, should be propagated correctly now :)

matthias-luger:
@Ekwav I played around with this for quite a while now and I notice very inconsistent behaivor.
I logged everytime I sent you a subFlip-Command and everytime I recieved a settingsChanged. But at times I just got random settingsChanged. Some of which are the default-values (minProfit: 0 etc.). I never sent such a subFlip.

Also: If I let it run in the background I get sent settingsChanged regularly. Some "defaut-values" and some with the correct ones: 
image


Are you sure you send it correctly? If so, do you have any idea what I could be doing wrong?

Ekwav:
I am sure that you only get a settings changed if your account sent an update.
I could be related that authorizing the mod also triggers an update.
Settings updates are not initiated by the backend in any form.
They are all written to a Kafka queue which is read by all programs that use your settings (that's how it's synced)

I used to check the settings against one stored in server cache but that's currently deactivated due to the recent instability issues.

matthias-luger:
@Ekwav I might have found the issue. I think it has something todo with the temporary workaround with the second websocket (for the old payment system). I would like to test this again, after the new payment system is merged (#626)

matthias-luger:
@matthias-luger remember to include the blockTenSecMsg-Setting after #636 is merged

matthias-luger:
next implementation #717

PR #610, opened 2021-12-04
maybe fix expired token?

matthias-luger:
This may fix the expired token bug


Ekwav (approved):
Can't really test this so let's give it a try

PR #612, opened 2021-12-04
fix scrolling & sold auction

### LINKED ISSUE #609: Clearing flips breaks the scroll event
After using the clear flips button, the mouse wheel scroll doesn't work anymore. I suspect the event handler doesn't get handled properly.

matthias-luger:
fixes #609


Ekwav (approved):
LGTM! :)

PR #615, opened 2021-12-04
use local socket when local

Ekwav:


src/properties.js
```diff
@@ -1,7 +1,7 @@
 let properties = {
     "commandEndpoint": window.location.host === 'localhost:3000' || window.location.hostname.includes("pr-env-sky-") ? "https://sky.coflnet.com/command" : "/command",
     "apiEndpoint": window.location.host === 'localhost:3000' || window.location.hostname.includes("pr-env-sky-") ? "https://sky.coflnet.com/api" : "/api",
-    "websocketEndpoint": "wss://sky.coflnet.com/skyblock",
+    "websocketEndpoint": window.location.host === 'localhost:8008' ? "ws://localhost:8008/skyblock" : "wss://sky.coflnet.com/skyblock",
```
matthias-luger, line 4: I never run the backend locally to develop the frontend. Wouldn't this break for me then?
Maybe add a way to toggle this

matthias-luger (changes requested):

src/properties.js
Ekwav: Why would it break for you if you don't use the backend?
It only changes the backend if your frontend is served by the local backend (port 8008) so doesn't affect you at all.

matthias-luger: You are right, sorry :)

PR #624, opened 2021-12-20
The player filter now sends the uuid

### LINKED ISSUE #590: Set player uuid for seller filter
The seller filter currently sets the player name. This leads to issues when players update their name.
The seller filter should always send the player uuid

matthias-luger:
fixes #590


Ekwav (approved):
LGTM!

PR #629, opened 2021-12-22
removed private properties

### LINKED ISSUE #625: Seller filter change breaks
The recent update replacing the seller name with seller uuid is sending the invalid filter `_hide` cause the filter to effectively break.
The key `_hide` should not be sent via `subFlip`

matthias-luger:
fixes #625


Ekwav (approved):
LGTM!! 🙃

PR #630, opened 2021-12-23
changed colour to hex

### LINKED ISSUE #628: Change color NBT data
Most users are used to seeing the color as hex.
Thus the NBT data "color" on auction details should display the hex value by default and the decimal value in a tooltip

matthias-luger:
fixes #628

### DIFF
--- a/src/components/AuctionDetails/AuctionDetails.tsx
+++ b/src/components/AuctionDetails/AuctionDetails.tsx
@@ -171,7 +171,7 @@ function AuctionDetails(props: Props) {
             decSplits.forEach(split => {
                 hexSplits.push(parseInt(split).toString(16));
             })
-            return <Tooltip type="hover" content={<span>{hexSplits.join(":")}</span>} tooltipContent={value} />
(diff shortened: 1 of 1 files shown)


src/components/AuctionDetails/AuctionDetails.tsx
```diff
@@ -164,6 +165,15 @@ function AuctionDetails(props: Props) {
             return value;
         }
 
+        if (key === "color") {
+            let decSplits = value ? value.split(":") : [];
+            let hexSplits: string[] = [];
+            decSplits.forEach(split => {
+                hexSplits.push(parseInt(split).toString(16));
+            })
+            return <Tooltip type="hover" content={<span>{hexSplits.join(":")}</span>} tooltipContent={value} />
```
Ekwav:
```suggestion
            return <Tooltip type="hover" content={<span>{hexSplits.join("")}</span>} tooltipContent={value} />
```
The hex format isn't seperated by colons, the parts are always the same length

Ekwav (changes requested):

PR #637, opened 2021-12-27
added maxInteger as defaultValue for maxCost

### LINKED ISSUE #635: Update max cost field
#634 didn't limiting max cost. 
People seem to be confused by the default value of `0`
Therefor the default should be set to the maximum value of int32 which should also be the maximum value that can be entered

matthias-luger:
fixes #635


Ekwav (approved):
LGTM! :)

PR #643, opened 2021-12-31
changed wording from subscription to

### LINKED ISSUE #640: Rename subscriptions to notifications
Users get confused by subscriptions

matthias-luger:
fixes #640


Ekwav (approved):
Not sure if notifiers is the best word but I can't think of a better one

PR #651, opened 2022-01-05
implemented player with rank filter

### LINKED ISSUE #647: Add custom handling for player with rank filter
Cake souls don't just include a player name but also a rank.
This rank is the one the player had when the cake soul was captured.
Users should be able to filter cake souls by captured player.
Therefore the filter type `512` marks any filter where the input should be a rank dropdown paired with a playersearch.
A list of all ranks can be found here: https://github.com/NotEnoughUpdates/NotEnoughUpdates-REPO/blob/df3a7de04d7a20e05d402b6df1c63db2b5e69091/constants/misc.json#L1039
common samples are
* §c[§fYOUTUBE§c]
* §6[MVP§f++§6]
* §b[MVP§c+§b]
* §c[ADMIN]

most ranks seem to postfix the name with `§f`
The output/input of the filter (serverside) should be a string consisting of `Rank` `Name` (with space between) 
There should two special options of ranks named `No player` (default with input disabled) and `no rank` (pre and postfixing the name with §7)

There may be other weird changes, we have to test this for a few names first.

matthias-luger:
fixes #647


Ekwav (changes requested):
The filter should be applied as soon as its added (the default curently is [no player] bit that is not useable because its not applied)
[no player] should not allow the user to search for a player, if possible clicking on the input while [no player] is selected trigger the dropdown or completely hide the input.
The player search should send the player name instead of the uuid.
This is not a direct player search but a player name search.
Cake souls keep the name of the player forever.

src/api/ApiHelper.tsx
```diff
@@ -77,6 +77,9 @@ function initAPI(): API {
                 start: Math.round(fetchStart / 100000) * 100,
                 filter: itemFilter
             };
+
+            console.log(requestData.filter);
```
Ekwav: please remove this

Ekwav (changes requested):
Can't be started 

./src/bootstrap-react.min.css
Error: PostCSS plugin postcss-purgecss requires PostCSS 8.
Migration guide for end-users:
https://github.com/postcss/postcss/wiki/PostCSS-8-for-end-users

matthias-luger:
@Ekwav it started for me. Maybe try npm install?

matthias-luger:
@Ekwav Please try again and do a "npm install"

Ekwav (approved):
looks good :)

PR #655, opened 2022-01-08
Boolean filter

### LINKED ISSUE #654: Add a drop-down to boolean filters
The filter type boolean can currently only be set to true.
There should be a drop-down added with the default value of true but the alternative value false.

matthias-luger:
fixes #654


matthias-luger:
based on #651 as this reworks the filter elements

Ekwav (changes requested):
please fix merge conflicts

Ekwav (approved):

PR #660, opened 2022-01-15
added new mod settings

### LINKED ISSUE #659: Add new settings
There are a few new settings.
Json keys:
* `mod.chat` enables the cofl chat by default
* `visibility.sellerOpenBtn` shows a button that opens the sellers ah
* `visibility.lore` shows the item lore when hovering chat
* `mod.format` Allows to change the mod format (is a string)

The default format is about the following:
"\n{0}: {1}{2} {3}{4} -> {5} (+{6} {7}) Med: {8} Lbin: {9} Volume: {10}"
                    {0} FlipFinder
                    {1} Item Rarity Color
                    {2} Item Name
                    {3} Price color
                    {4} Starting bid
                    {5} Target Price
                    {6} Estimated Profit
                    {7} Provit percentage
                    {8} Median Price
                    {9} Lowest Bin
                    {10}Volume
Every $ is replaced with § (formating prefix for minecraft)
There should be an editor which renders a message preview with the color library.

matthias-luger:
fixes #659

### DIFF
--- a/src/api/ApiHelper.tsx
+++ b/src/api/ApiHelper.tsx
@@ -524,7 +524,9 @@ function initAPI(): API {
                 justProfit: flipSettings.justProfit,
                 soundOnFlip: flipSettings.soundOnFlip,
                 shortNumbers: flipSettings.shortNumbers,
-                blockTenSecMsg: flipSettings.blockTenSecMsg
+                blockTenSecMsg: flipSettings.blockTenSecMsg,
+                format: flipSettings.modFormat,
+                chat: !flipSettings.hideModChat
             },
             visibility: {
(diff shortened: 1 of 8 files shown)


Ekwav (changes requested):
The format resets after every reload and it can't be deactivated.
The custom format should be an opt in for the user. 
Ie it should have no value by default a button or similar to get the default format and another button to deactivate the custom format (by setting it to null or empty string again)

Ekwav (changes requested):
clicking the format reset button doesn't reset the text in the input field. It looks like that is it not how it is intended to us.

Ekwav (approved):
LGTM! :)

PR #665, opened 2022-01-25
tiny changes for next

matthias-luger:
using next to implement SSR


LICENSE
```diff
@@ -1,661 +0,0 @@
-                    GNU AFFERO GENERAL PUBLIC LICENSE
```
Ekwav, line 1: Why was the licence removed?

Ekwav (changes requested):

PR #669, opened 2022-01-31
added defaultValue parameter to getSettings

matthias-luger:


Ekwav:
What about already exported and broken filters?

Ekwav (approved):

PR #675, opened 2022-02-01
changed select to typeahead

### LINKED ISSUE #671: Change filter select to input
There are a lot of filters. Therefore the filter select box should be changed to be searchable like enchantments.

matthias-luger:
fixes #671


Ekwav (commented):
Why does this change reformat 300 lines?

matthias-luger:
Because formatting...

PR #684, opened 2022-02-15
fixed endless google loading

matthias-luger:


Ekwav (commented):
was told this is the cause, lets see

PR #701, opened 2022-02-28
added median, volume and slayer collection to crafts

### LINKED ISSUE #698: Add slayer requirements to profitable crafts
Some crafts require a specific slayer type and level instead of a collection.
The additional property `reqSlayer` was added to map that dependency. It has the same format as `reqCollection` and should be displayed the same way but labeled as `Required Slayer`

### LINKED ISSUE #700: Add volume and median price to /crafts
The crafts data has been extended to include sell`volume` and `median`-price.
Both should be displayed to give users more context about how good a craft flip is.

matthias-luger:
fixes #700
fixes #698

### DIFF
--- a/src/components/CraftsList/CraftsList.css
+++ b/src/components/CraftsList/CraftsList.css
@@ -35,3 +35,12 @@
     -ms-user-select: none;
     width: calc(50% - 10px);
 }
+
+.crafts-list .label {
+    width: 150px;
+    float: left;
+}
+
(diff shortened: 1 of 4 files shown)


Ekwav (changes requested):
There should only be the one item requirement line.
There are no items that require both a `slayer` and `collection` level

Ekwav (approved):

PR #717, opened 2022-04-05
Next server settings update

### LINKED ISSUE #679: Load settings from the server
Flip Settings are now stored server side. They can and should be retrieved from the server with either `getFlipSettings` or by passing the serialized form of `null` to `subFlip`.
Both of which will respond with the current settings. The setting value may be null as well in which case the current settings should be sent to `subFlip`.
Every update should continue to use `subFlip` as well

matthias-luger:
fixes #679
This is based on https://github.com/Coflnet/hypixel-react/pull/665

### DIFF
--- a/components/Flipper/FlipperFilter/FlipperFilter.tsx
+++ b/components/Flipper/FlipperFilter/FlipperFilter.tsx
@@ -124,7 +124,7 @@ function FlipperFilter(props: Props) {
         setOnlyUnsold(isActive)
         let filter = getCurrentFilter()
         filter.onlyUnsold = isActive
-        api.setFlipSetting('maxCost', isActive)
+        api.setFlipSetting('showHideSold', isActive)
         onFilterChange(filter)
     }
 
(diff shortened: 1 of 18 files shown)


matthias-luger:
This is based on #665

matthias-luger:
@Ekwav The backend needs to implement a new property called "copySuccessMessage". This is a setting only relevant to the UI therefore the backend didnt need to know about it until know.
I would think it should belong under the "visibility"-Property

matthias-luger:
@Ekwav There is also a UI Property I send you called "links". I put it into the visibility-Property as well

utils/SettingsUtils.tsx
```diff
@@ -49,6 +53,122 @@ export function setSetting(key: any, value: any) {
...
+        )
+
+        setSetting(
+            FLIPPER_FILTER_KEY,
+            JSON.stringify({
+                maxCost: data.maxCost,
+                minProfit: data.minProfit,
+                minProfitPercent: data.minProfitPercent,
+                minVolume: data.minVolume,
+                onlyBin: data.filters.Bin === 'true'
```
Ekwav: as you requested this got changed to 
```suggestion
                onlyBin: data.onlyBin
```

components/Flipper/FlipperFilter/FlipperFilter.tsx
```diff
@@ -116,13 +116,15 @@ function FlipperFilter(props: Props) {
...
         filter.maxCost = val
+        api.setFlipSetting('maxCost', val)
         onFilterChange(filter)
     }
 
     function updateOnlyUnsold(isActive: boolean) {
         setOnlyUnsold(isActive)
         let filter = getCurrentFilter()
         filter.onlyUnsold = isActive
+        api.setFlipSetting('maxCost', isActive)
```
Ekwav: this should map to `showHideSold`

Ekwav (changes requested):
OnlyBin and hideSold don't work

Ekwav (approved):

PR #723, opened 2022-04-07
readded missing flip settings

### LINKED ISSUE #722: The flip settings are not all included
subFlip is at least missing sellerButton and customFormat.
maybe more

matthias-luger:
fixes #722


Ekwav (approved):
Couldn't test it but also couldn't find any obvious mistake

PR #729, opened 2022-04-11
set a bin true filter for whitelist items

### LINKED ISSUE #716: Add Bin to whitelisted items
> Default BIN to true in whitelisted items unless specifically changed to false

by vytal
Whitelisted items are likely only of interest if they can be bought directly

matthias-luger:
fixes #716


Flou21 (approved):
lgtm

PR #731, opened 2022-04-11
Implemented Incremental Static Regeneration

### LINKED ISSUE #712: Preload pages on hover
To make the website even more responsive links should be prefetched when hovered (either any hover or after a cooldown)
Ie. not only the required javascript but the api requests for needed data should be triggered as well.
This looked promising
https://www.webperf.tips/tip/hover-preloading/

matthias-luger:
fixes #712


matthias-luger:
Further testing in terms of server capacity is required

Flou21 (approved):
lgtm

PR #736, opened 2022-04-14
added button to manually trigger the endless loading

### LINKED ISSUE #706: Endless scroll breaks on big screens
The endless scroll depends on the scroll event being triggered.
However on 4k screens the whole list of auctions or bids fits on screen which results in the user being unable to scroll.

Please also switch from /commands to using the /api while modifying the players page.

matthias-luger:
fixes #706


Ekwav (approved):
LGTM! :)
Doesn't load after scrolling started tho (the button has to be pressed for each new batch)

PR #743, opened 2022-04-16
moved ref system to api

matthias-luger:


matthias-luger:
Need to be tested

api/ApiHelper.tsx
```diff
@@ -926,9 +925,19 @@ export function initAPI(returnSSRResponse: boolean = false): API {
...
+            }
+
+            httpApi.sendApiRequest({
                 type: RequestType.GET_REF_INFO,
                 data: '',
+                requestHeader: {
+                    GoogleToken: googleId
+                },
                 resolve: (response: any) => {
                     resolve(parseRefInfo(response))
```
Ekwav, line 942: the format changed

api/ApiHelper.tsx
```diff
@@ -926,9 +925,19 @@ export function initAPI(returnSSRResponse: boolean = false): API {
 
     let getRefInfo = (): Promise<RefInfo> => {
         return new Promise((resolve, reject) => {
-            websocketHelper.sendRequest({
+            let googleId = localStorage.getItem('googleId')
+            if (!googleId) {
+                toast.error('You need to be logged in to use the ref system something.')
```
Ekwav:
```suggestion
                toast.error('You need to be logged in to use the ref system.')
```

api/ApiHelper.tsx
```diff
@@ -942,17 +951,33 @@ export function initAPI(returnSSRResponse: boolean = false): API {
...
     let setRef = (refId: string): Promise<void> => {
         return new Promise((resolve, reject) => {
-            websocketHelper.sendRequest({
-                type: RequestType.SET_REF,
-                data: refId,
-                resolve: () => {
-                    resolve()
+            let googleId = localStorage.getItem('googleId')
+            if (!googleId) {
+                toast.error('You need to be logged in to use the ref system something.')
```
Ekwav: I don't think showing an error to new users is a good idea.

Ekwav (changes requested):
Neither requests work, setRef isn't even called according to the network tab

api/ApiHelper.tsx
matthias-luger: This should really never occur as the function is called in the onAfterLogin.
But if for some reason it is called before, why should I not send out an error? What should I do instead?

Ekwav (approved):

PR #748, opened 2022-04-17
use percice time returned

Ekwav:
I unified the time to be calculated not based on the time hypixel says but the time I actually was able to load new auctions.
This should fix any out of sync issues that were reported


matthias-luger (approved):
LGTM

PR #752, opened 2022-04-19
reject if a failed http response is received

matthias-luger:


Ekwav (approved):
Sounds like a good idea 🙂

PR #753, opened 2022-04-19
implemented button to reset settings

### LINKED ISSUE #692: Reset to defaults button
Add a reset to defaults button in the settings menu.

A confirmation prompt would be reasonable. So you do not accidentally delete your blacklist.

matthias-luger:
fixes #692

### DIFF
--- a/components/Flipper/Flipper.tsx
+++ b/components/Flipper/Flipper.tsx
@@ -631,7 +631,7 @@ function Flipper(props: Props) {
                     {isLoggedIn ? (
                         ''
                     ) : (
-                        <>
+                        <div>
                             <span>
                                 These are flips that were previosly found (~5 min ago). Anyone can use these and there is no cap on estimated profit. Keep in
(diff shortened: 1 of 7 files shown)


components/Flipper/Flipper.tsx
```diff
@@ -563,14 +631,18 @@ function Flipper(props: Props) {
...
                         ''
                     ) : (
-                        <span>
-                            These are flips that were previosly found (~5 min ago). Anyone can use these and there is no cap on estimated profit. Keep in mind
-                            that these are delayed to protect our paying supporters. If you want more recent flips purchase our{' '}
-                            <a target="_blank" rel="noreferrer" href="/premium">
-                                premium plan.
-                            </a>
-                        </span>
+                        <>
```
Ekwav: what the heck is that

matthias-luger: Thats a Fragment (https://reactjs.org/docs/fragments.html#short-syntax)
It's a nice way to group multiple elements in one component without having to group it into a useless div

PR #760, opened 2022-04-21
fixed usage of the properties of the old ref system

matthias-luger:


matthias-luger:
@Ekwav The build fails without this fix

Ekwav (approved):

PR #764, opened 2022-04-21
add missing revalidate

### LINKED ISSUE #763: New auctions, new items & more aren't updated
The start page is not updated with new data for at least 
- new auctions
- new players
- new items
I suspect popular searches as well although that's not as important.

matthias-luger:
fixes #763


Ekwav (approved):
Not sure if that works that way but I hope so

PR #766, opened 2022-04-22
allow settings in the toml format to be imported

matthias-luger:


components/Flipper/Flipper.tsx
```diff
@@ -335,6 +335,8 @@ function Flipper(props: Props) {
     }
 
     function addItemToBlacklist(flip: FlipAuction) {
+        console.log(flip.props)
```
Ekwav: Should be removed

pages/flipper.tsx
```diff
@@ -1,19 +1,58 @@
...
     flips?: any[]
 }
 
 function Flipper(props: Props) {
+    let dragCounter = useRef(0)
+
+    function onDragStart(e) {
+        e.preventDefault()
+        dragCounter.current++
+        console.log('on drag start')
```
Ekwav: Should be removed

pages/flipper.tsx
```diff
@@ -1,19 +1,58 @@
...
+    }
+
+    function onDragEnd(e) {
+        e.preventDefault()
+        dragCounter.current--
+        if (dragCounter.current === 0) {
+            console.log('on drag end')
```
Ekwav: Should be removed

Ekwav (changes requested):

Ekwav (changes requested):
Doesn't accept all options eg user blacklist

PR #774, opened 2022-04-26
remove lazy loading for search

matthias-luger:
I want to remove the lazy loading for the search.
Sometimes it takes too long until the browser loads the images of the search, resulting in a search that looks "broken" or "slow".

It looks much better that way :)


Ekwav (approved):
Makes sense since the search is basically always in view

PR #780, opened 2022-04-28
you can now edit restrictions

### LINKED ISSUE #710: Make blacklist/whitelist editable
Add a option to edit existing blacklist/whitelist entries.

matthias-luger:
fixes #710


components/Flipper/FlipRestrictionList/EditRestriction/EditRestriction.tsx
```diff
@@ -0,0 +1,83 @@
...
+                            }}
+                        >
+                            Add
+                        </Button>
+                    }
+                    tooltipContent={
+                        <span>
+                            Adds the filter to the selected restriction(s). If a specific filter is already present, it is <b>not</b> overwritten.
+                        </span>
+                    }
```
Ekwav: when selecting one filter this button does nothing when being clicked.
It should either be hidden or displaying a note why it does nothing. 
I got very confused by it

components/Flipper/FlipRestrictionList/EditRestriction/EditRestriction.tsx
```diff
@@ -0,0 +1,83 @@
...
+                            style={{ marginLeft: '20px' }}
+                        >
+                            Add (override existing)
+                        </Button>
+                    }
+                    tooltipContent={
+                        <span>
+                            Adds the filter to the selected restriction(s). If a specific filter is already present, it <b>is overwritten</b>.
+                        </span>
+                    }
```
Ekwav: same for this.
also when removing a filter the add can be clicked but does nothing which I found to be pretty frustrating.  
Maybe make it live preview the changes after the update to avoid confusion

Ekwav (changes requested):
I am missing a cancle Edit button at the top and the ability to change black to whitelist entries.
Also I tried to edit one entry with 3 filters and only got BIN in the select 
image

Ekwav (changes requested):
image
(I didn't do anything except clicking the edit button)

- I am missing a filter copy function
- There is no way to switch black<->whitelist  
- the "add" button removes filters?
- also generally why can't I just edit one list entry in place and have to use the bulk edit for everything. I am pretty sure users commonly only edit one entry

components/ItemFilter/ItemFilterPropertiesDisplay.tsx
```diff
@@ -15,16 +18,16 @@ const SELLER_FORMAT_FILTER = 'Seller'
...
-            return;
+        if (!props.filter) {
+            return
         }
-        let localFilter = JSON.parse(JSON.stringify(props.filter));
+        let localFilter = JSON.parse(JSON.stringify(props.filter))
         setLocalFilter(localFilter)
         checkForSellerName(localFilter)
-    }, [props.filter])
+    }, [JSON.stringify(props.filter)])
```
Ekwav: please optimise this

Ekwav (changes requested):

Ekwav (approved):

PR #783, opened 2022-04-28
added account page

matthias-luger:


components/AccountDetails/AccountDetails.tsx
```diff
@@ -0,0 +1,251 @@
...
+                        buttonText="Logout"
+                        onLogoutSuccess={onLogout}
+                    />
+                </div>
+            ) : null}
+            <hr />
+            <h2 style={{ marginBottom: '30px' }}>Settings</h2>
+            <p>
+                <div style={{ display: 'inline-block' }}>
+                    <span className={styles.label}>Allow cookies for tracking: </span>
```
Ekwav:
```suggestion
                    <span className={styles.label}>Allow cookies for analytics: </span>
```
We don't track people around the web, we only use matomo analytics

Ekwav (changes requested):

Ekwav (approved):

PR #784, opened 2022-04-29
fix concatination of url params

matthias-luger:


api/HttpHelper.tsx
```diff
@@ -69,6 +69,9 @@ export function initHttpHelper(customCommandEndpoint?: string, customApiEndpoint
         }
 
         if (cacheInvalidationGrouping) {
+            if (url.charAt(url.length - 1) === '&' || url.charAt(url.length - 1) === '?') {
+                url = url.substring(0, url.length - 1)
+            }
             if (url.indexOf('?') !== -1) {
                 url += `&t=${cacheInvalidationGrouping}`
             } else {
```
Ekwav: this can break api endpoints that take a filter query. Are you aware of that and is this on purpose?

Ekwav (approved):

PR #797, opened 2022-05-10
- optimized import to not request all the item names

matthias-luger:
- fixed a problem with the base64 encoding of utf8 strings
- refactored the sub-commands to remove duplicate code


Ekwav (approved):
LGTM! :)

PR #801, opened 2022-05-12
store last used filter

### LINKED ISSUE #745: Remember filters between sessions
The filters should be remembered client side between sessions. Users commonly use the same filters and this would remove the need to reenter the filter values and be just one extra click to remove them in case they actually don't want them.

matthias-luger:
fixes #745


Ekwav (changes requested):
Doesn't work for me.
I searched up Adaptive leggings, added a rarity filter, closed the page, opened it again, and searched for adaptive boots, no filters there

Ekwav (changes requested):
Mostly works now :+1: , but the filter select opens when there already is a filter. It should only open when there is none.

Ekwav (approved):

PR #802, opened 2022-05-12
redirect to startpage if user comes from google to search specific pl…

matthias-luger:
…ayer

### DIFF
--- a/components/Hyauctions/Hyauctions.tsx
+++ b/components/Hyauctions/Hyauctions.tsx
@@ -6,7 +6,7 @@ function Hyauctions() {
             <h2>Coflnet</h2>
             <hr />
             <p>We provide a hyauctions alternative. That includes player and item history but also bazaar history, notifications, craft & kat flips as well as an advanced ah flipper.</p>
-            <p>Just try using the searchbar above.</p>
+            <p>Just try using the search bar above. (it actually works)</p>
         </>
     )
 }
--- a/components/Hyauctions/Hyauctions.tsx
+++ b/components/Hyauctions/Hyauctions.tsx
@@ -1,3 +1,4 @@
+import Link from 'next/link'
 import React from 'react'
 
 function Hyauctions() {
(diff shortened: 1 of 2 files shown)


pages/player/[uuid]/index.tsx
```diff
@@ -37,6 +37,11 @@ function PlayerDetails(props: Props) {
     let removeSwipeListeners = useSwipe(undefined, onSwipeRight, undefined, onSwipeLeft)
 
     useEffect(() => {
+        // special case for people searching hyauctions
+        if (window.document.referrer.includes('google') && uuid === 'be7002531956406d81c535a81fe2833a') {
+            router.push('/')
+            return
+        }
```
Comment (Ekwav) [inline pages/player/[uuid]/index.tsx]: there should be some special text on page/ or a dedicated `/hyauctions` page detailing why its not as good as ours, just redirecting is not a good idea

Ekwav (changes requested):

components/Hyauctions/Hyauctions.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react'
+
+function Hyauctions() {
+    return (
+        <>
+            <h2>Coflnet</h2>
+            <hr />
+            <p>Seems you were looking for Hyauctions. We also provide a auction history to check prices and player auctions.</p>
```
Ekwav:
```suggestion
            <p>We provide a hyauctions alternative. That includes player and item history but also bazaar history, notifications, craft & kat flips as well as an advanced ah flipper.</p>
```

components/Hyauctions/Hyauctions.tsx
```diff
@@ -0,0 +1,14 @@
...
+            <p>Just try using the searchbar above.</p>
```
Ekwav:
```suggestion
            <p>Just try using the search bar above. (it actually works)</p>
```
poke fun at the broken search of hyauctions

Ekwav (changes requested):

components/Hyauctions/Hyauctions.tsx
```diff
@@ -0,0 +1,18 @@
...
+function Hyauctions() {
+    return (
+        <>
+            <h2>Coflnet</h2>
+            <hr />
+            <p>
+                We provide a hyauctions alternative. That includes player and item history but also bazaar history, notifications,{' '}
+                <Link href={'/crafts'}>crafts</Link> & <Link href={'/kat'}>kat flips</Link> as well as an <Link href={'/flipper'}>advanced ah flipper</Link>.
+            </p>
+            <p>Just try using the search bar above (it actually works).</p>
```
matthias-luger, line 13: @Ekwav The "it actually works" sounds kinda weird. Why do we need to emphasize that a feature works?

Ekwav: because hyauctions is well know  for that their search doesn't work/ doesn't find things

Ekwav (approved):

PR #806, opened 2022-05-16
- added tooltip to disabled payment buttons

### LINKED ISSUE #804: Highlight accepting the no return policy
Our users get confused because by default buy cofl coins button is not clickable.
Either move the checkbox for accepting that they wave their return right into a modal or highlight it some other way.

matthias-luger:
fixes #804

### DIFF
--- a/components/Payment/Payment.tsx
+++ b/components/Payment/Payment.tsx
@@ -34,6 +34,26 @@ function Payment(props: Props) {
         })
     }
 
+    function getDisabledPaymentTooltip() {
+        return !props.cancellationRightLossConfirmed ? <span>Please note the information regarding your cancellation right above.</span> : null
+    }
+
+    function getRoundedPrice(price: number) {
+        return Math.round(price * 100) / 100
+    }
+
+    let paypalHigherPricesTooltip = (
+        <Tooltip
+            content={
+                <span style={{ marginLeft: '5px' }}>
+                    <HelpIcon />
+                </span>
+            }
+            type="hover"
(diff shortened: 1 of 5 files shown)


components/Payment/Payment.tsx
```diff
@@ -54,50 +54,70 @@ function Payment(props: Props) {
...
-                                </p>
-                            ) : (
-                                `${numberWithThousandsSeperators(Math.round(stripePrice * 100) / 100)} Euro`
-                            )}
-                        </Button>
+                        <div className={styles.paymentLabel}>Buy with Stripe</div>
+                        <Tooltip
+                            type="hover"
+                            tooltipContent={
+                                !props.cancellationRightLossConfirmed ? <span>Please note the information regarding your cancellation right above.</span> : null
```
Ekwav: the message content (and check) is duplicated, please refactor it so it only exists once

components/Payment/Payment.tsx
```diff
@@ -54,50 +54,70 @@ function Payment(props: Props) {
...
+                                            <p className={styles.manualRedirectLink}>
+                                                Redirecting to Stripe...
+                                                <br />
+                                                Not working?{' '}
+                                                <a href={currentRedirectLink} target="_blank">
+                                                    Click here
+                                                </a>
+                                            </p>
+                                        ) : (
+                                            `${numberWithThousandsSeperators(Math.round(stripePrice * 100) / 100)} Euro`
```
Ekwav: calculation is also dupplicated

Ekwav (changes requested):

Ekwav (approved):

PR #809, opened 2022-05-17
show useful information about the ref system before login

matthias-luger:


matthias-luger:
@Flou21 You said, you want to review that?

PR #840, opened 2022-06-21
added error message if flipBased fails

### LINKED ISSUE #833: Similar sells loads forever
The request fails after 246ms but the loading stays forever
image
It should either display an error message or retry

matthias-luger:
fixes #833

### DIFF
--- a/components/Flipper/FlipBased/FlipBased.tsx
+++ b/components/Flipper/FlipBased/FlipBased.tsx
@@ -16,14 +16,20 @@ interface Props {
 function FlipBased(props: Props) {
     let [auctions, setAuctions] = useState<Auction[]>([])
     let [isLoading, setIsLoading] = useState(true)
+    let [hasLoadingFailed, setHasLoadingFailed] = useState(false)
 
     let forceUpdate = useForceUpdate()
 
     useEffect(() => {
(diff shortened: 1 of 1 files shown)


components/Flipper/FlipBased/FlipBased.tsx
```diff
@@ -78,13 +84,12 @@ function FlipBased(props: Props) {
...
-            ) : auctions.length > 0 ? (
+            {isLoading ? getLoadingElement() : null}
+            {!isLoading && !hasLoadingFailed && auctions.length > 0 ? (
                 <div style={{ display: 'flex', flexWrap: 'wrap', justifyContent: 'center', alignItems: 'stretch' }}>{auctionsElement}</div>
-            ) : (
-                <p>No auctions found</p>
-            )}
+            ) : null}
+            {!isLoading && !hasLoadingFailed && auctions.length === 0 ? <p>No auctions found</p> : null}
+            {hasLoadingFailed ? <p>An error occured while loading the auctions...</p> : null}
```
Ekwav:
```suggestion
            {hasLoadingFailed ? <p>An error occured while loading the auctions</p> : null}
```
the ... makes it look like there will something happen/more info to come but it isn't

Ekwav (changes requested):

PR #843, opened 2022-06-21
added option to create a notifier if a player creates an auction

### LINKED ISSUE #827: Add notification for player auction creation
There is no checkbox for player notifications to get a notification when a player creates and auction.
Please add an extra checkbox, the enum type for that kind of notification is 128

matthias-luger:
fixes #827

### DIFF
--- a/components/SubscriptionList/SubscriptionList.tsx
+++ b/components/SubscriptionList/SubscriptionList.tsx
@@ -88,6 +88,9 @@ function SubscriptionList() {
                         case SubscriptionType.SOLD.toString():
                             result = <li key="5">Notify if player sold something</li>
                             break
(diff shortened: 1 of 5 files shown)


components/SubscriptionList/SubscriptionList.tsx
```diff
@@ -127,6 +130,7 @@ function SubscriptionList() {
     }
 
     function getSubscriptionTitle(subscription: Subscription): Promise<void> {
+        console.log(subscription)
```
Ekwav: please remove

Ekwav (changes requested):

Ekwav (approved):

PR #844, opened 2022-06-22
add delay tooltip for seller option

### LINKED ISSUE #826: tooltip for seller flipper option
add a tooltip to the seller flipper setting, warning the user that this adds a delay

to the other options that can create a delay as well

matthias-luger:
fixes #826


Flou21:
the tooltip does not show when you hover over the checkbox
it only works when you hove over the "Seller" Text

a ? icon would be nice, like "use lowest bin to calc"

Flou21 (changes requested):
the tooltip does not show when you hover over the checkbox
it only works when you hove over the "Seller" Text

a ? icon would be nice, like "use lowest bin to calc"

Flou21 (approved):
lgtm

PR #847, opened 2022-06-22
added error handling for payment redirects

matthias-luger:


components/Payment/Payment.tsx
```diff
@@ -20,18 +21,30 @@ function Payment(props: Props) {
...
-        })
+        api.paypalPurchase(productId, coflCoins)
+            .then(data => {
+                setCurrentRedirectLink(data.directLink)
+                window.open(data.directLink)
+            })
+            .catch(() => {
+                setCurrentRedirectLink('')
+                setLoadingId('')
+                toast.error('Something went wrong. Please try again.')
```
Ekwav: any mention of traceId or the error message if its not "internal_error"?

Ekwav (changes requested):

components/Payment/Payment.tsx
Ekwav: also, redundant code

matthias-luger: > any mention of traceId or the error message if its not "internal_error"?

If I get a error object from the backend, I display the message. This is the default error handling in the apihelper.
image

Ekwav (approved):

PR #850, opened 2022-06-23
added range slider element for enchantment level filter

### LINKED ISSUE #693: Change the enchantment level select to radio
Change the enchantment level select to a radio select.
It should be able for the user to select ranges of levels with one field

matthias-luger:
fixes #693


Ekwav (changes requested):
Doesn't work from URL
Breaks existing filters such as `>5` `<5`
requires two clicks and two drags of slider just to select lvl 6, could you optimise that?
Slider overlaps with other elements, it should have a bit of right and left padding

Ekwav (changes requested):
works as expected :+1: 
The input field could be made smaller tho
image

matthias-luger:
The field was already smaller than that, I now made it even smaller though...
image
image

Ekwav (approved):

PR #853, opened 2022-06-24
Fix endless load restriction list

matthias-luger:


matthias-luger:
@Ekwav I reverted back to the stringify solution. The deconstruction always triggered, as the deconstructed properties apparently are always counted as different (similar to new instances of objects I suppose). 

I checked and everyone of these is only called once if the restrictions are opened and if the filter is changed (as it should), so the performance issue should be minimal.

Ekwav (approved):
Ok

Issue #864, opened 2022-07-09
Bids don't refresh when navigating 


Ekwav:
<img width="1118" alt="image" src="https://user-images.githubusercontent.com/16632490/178114333-aa43cc51-47a1-419e-bb68-3c1c52ff95aa.png">
<img width="1074" alt="image" src="https://user-images.githubusercontent.com/16632490/178114354-321dc3a4-2ba3-4084-b02c-f54a6d8761ff.png">

Ekwav:
I used the search to switch between the players.
Same issue for auctions

Ekwav:
the auctions also don't refresh for me.  
AND another report claimed wildly different tracked flips results

ThomasWu931:
I took a look at some of the files and I think the error  is caused because `PlayerDetailsList` doesn't unmount and remount when searching a new player. 

Since the `detailType` doesn't immediately change when swapping between players, `PlayerDetailsList` doesn't unmount, remount, and allow for `useState` to reinitialize the data. (https://github.com/Coflnet/hypixel-react/blob/develop/pages/player/%5Buuid%5D/index.tsx#L169-L176)

Flou21:
the issue was reported again by one of our users
please try to merge a fix asap

PR #883, opened 2022-07-31
Addressing "Bids don't refresh when navigating" (#864)

ThomasWu931:
### Issue

Currently, the bids and auctions sections don't change when searching between players. After switching to another player, old data for the previous player is kept on screen and new data isn't immediately loaded in.
More info in: https://github.com/Coflnet/hypixel-react/issues/864

### Solution

The issue seems to have come from the fact that `PlayerDetailsList` doesn't unmount and remount when searching between players. When searching between players,`detailType` remains unchanged causing change detection to not unmount `PlayerDetailsList` (https://github.com/Coflnet/hypixel-react/blob/develop/pages/player/%5Buuid%5D/index.tsx#L169-L176). This causes `useState` to not reinitialize `listElements` (https://github.com/Coflnet/hypixel-react/blob/develop/components/PlayerDetailsList/PlayerDetailsList.tsx#L40).

One solution can be to create a `useEffect` to call `setListElements` and some other fns whenever `props.UUID` changes

### Testing

NA

### Additional Notes

I also noticed that onRouteChange only has access to the initial value of `props`. I believe this is caused because `router.events.on('routeChangeStart', onRouteChange)` is only called once due to the empty second parameter of `useEffect`.(https://github.com/Coflnet/hypixel-react/blob/develop/components/PlayerDetailsList/PlayerDetailsList.tsx#L47-L70) This causes the event-listener to only call the onRouteChange  with the initial value of `props`.

More info [here](https://stackoverflow.com/questions/59881386/why-does-the-callback-on-addevenetlistener-print-out-old-state): 

In addition, I also felt that it would be better to remove auctions from the props of `PlayerDetailsList` and instead query and cache all of the player data within listStates. This is just to have a more consistent source of data. However, I'm not fully aware of optimizations that next.js uses so this could be a bad choice.

This PR also addressed thos eissues

### DIFF
--- a/components/PlayerDetailsList/PlayerDetailsList.tsx
+++ b/components/PlayerDetailsList/PlayerDetailsList.tsx
@@ -14,9 +14,9 @@ import styles from './PlayerDetailsList.module.css'
 
 interface Props {
     playerUUID: string
+    playerName: string
     loadingDataFunction: Function
     type: 'auctions' | 'bids'
-    auctions?: Auction[]
 }
 
 interface ListState {
@@ -43,51 +43,44 @@ function PlayerDetailsList(props: Props) {
     let isLoadingElements = useRef(false)
 
     useEffect(() => {
+
         mounted = true
         router.events.on('routeChangeStart', onRouteChange)
 
+        setPlayerName(props.playerName)
+
         let listState = getListState()
-        if (listState !== undefined) {
+        if (listState) {
             setListElements(listState.listElements)
             setAllElementsLoaded(listState.allElementsLoaded)
-            setTimeout(() => {
-                if (!mounted) {
-                    return
-                }
-                window.scrollTo({
-                    left: 0,
-                    top: listState!.yOffset,
-                    behavior: 'auto'
-                })
-            }, 100)
+            isLoadingElements.current = false
         }
+        else {
+            // We need to set reset to true to remove the previous player's listElements
+            loadNewElements(true)
+        }
+
+        setTimeout(() => {
+            if (!mounted) {
+                return
+            }
+            window.scrollTo({
+                left: 0,
+                top: listState!.yOffset,
+                behavior: 'auto'
+            })
+        }, 100)
 
         return () => {
             mounted = false
             router.events.off('routeChangeStart', onRouteChange)
         }
-        // eslint-disable-next-line react-hooks/exhaustive-deps
-    }, [])
 
-    useEffect(() => {
-        api.getPlayerName(props.playerUUID).then(name => {
-            if (!mounted) {
-                return
-            }
-            setPlayerName(name)
-            setPlayerName(name)
-            setListElements(props.auctions || [])
-            setAllElementsLoaded(props.auctions ? props.auctions.length < 12 : false)
-        })
     }, [props.playerUUID])
 
-    useEffect(() => {
-        if (!props.auctions) {
-            loadNewElements()
-        }
-    }, [props.auctions])
 
     let onRouteChange = () => {
+        console.log("list States", listStates)
         let listState = getListState()
         if (listState) {
             listState.yOffset = window.pageYOffset
--- a/components/PlayerDetailsList/PlayerDetailsList.tsx
+++ b/components/PlayerDetailsList/PlayerDetailsList.tsx
@@ -37,9 +37,9 @@ function PlayerDetailsList(props: Props) {
     let forceUpdate = useForceUpdate()
     let router = useRouter()
 
-    let [listElements, setListElements] = useState<(Auction | BidForList)[]>(props.auctions || [])
-    let [allElementsLoaded, setAllElementsLoaded] = useState(props.auctions ? props.auctions.length < 12 : false)
-    let [playerName, setPlayerName] = useState('')
+    let [listElements, setListElements] = useState<(Auction | BidForList)[]>([])
+    let [allElementsLoaded, setAllElementsLoaded] = useState<boolean>()
+    let [playerName, setPlayerName] = useState<string>('')
     let isLoadingElements = useRef(false)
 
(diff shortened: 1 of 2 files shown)


Flou21:
@matthias-luger can you review that pr?

components/PlayerDetailsList/PlayerDetailsList.tsx
```diff
@@ -37,54 +37,50 @@ function PlayerDetailsList(props: Props) {
...
     }, [props.playerUUID])
 
-    useEffect(() => {
-        if (!props.auctions) {
-            loadNewElements()
-        }
-    }, [props.auctions])
 
     let onRouteChange = () => {
+        console.log("list States", listStates)
```
matthias-luger, line 83: remove the console.log

components/PlayerDetailsList/PlayerDetailsList.tsx
```diff
@@ -14,9 +14,9 @@ import styles from './PlayerDetailsList.module.css'
 
 interface Props {
     playerUUID: string
+    playerName: string
     loadingDataFunction: Function
     type: 'auctions' | 'bids'
-    auctions?: Auction[]
```
matthias-luger, line 19: please re-add the auctions here.
We want to have the auctions loaded serverside.

- We render the page server side => The auctions are already rendered, which is good for multiple SEO points
- We try to get as much initial data as possible from the server as the entire pre-rendered page is then cached for all users

components/PlayerDetailsList/PlayerDetailsList.tsx
```diff
@@ -37,54 +37,50 @@ function PlayerDetailsList(props: Props) {
...
-                    left: 0,
-                    top: listState!.yOffset,
-                    behavior: 'auto'
-                })
-            }, 100)
+            isLoadingElements.current = false
+        }
+        else {
+            // We need to set reset to true to remove the previous player's listElements
+            loadNewElements(true)
```
matthias-luger, line 60: As we preload the auctions again, this could have unwanted site effects.
You change the state here, to trigger a rerender, right?

There is a useful hook (`useForceUpdate`) which can be used here

matthias-luger, line 60: As we preload the auctions again, this could have unwanted site effects.
You change the state here, to trigger a rerender, right?

There is a useful hook (`useForceUpdate`) which can be used here

components/PlayerDetailsList/PlayerDetailsList.tsx
```diff
@@ -37,54 +37,50 @@ function PlayerDetailsList(props: Props) {
...
+            if (!mounted) {
+                return
+            }
+            window.scrollTo({
+                left: 0,
+                top: listState!.yOffset,
+                behavior: 'auto'
+            })
+        }, 100)
+
```
matthias-luger, line 73: Why did you move this outside the if?
It now throws a exception, because `listState` is undefined and we try to read ``yOffset` of undefined
image

pages/player/[uuid]/index.tsx
```diff
@@ -48,15 +46,8 @@ function PlayerDetails(props: Props) {
         if (!isClientSideRendering()) {
             return
         }
-        api.getPlayerName(uuid).then(name => {
```
Comment (matthias-luger) [inline pages/player/[uuid]/index.tsx:51]: We still need this refresh of the playername.
If the uuid changes through a shallow navigation (we are still on the same page, but the parameter in the url changed), the page is not rendered server side again, therefore props.player still includes the wrong one.

While this may lead to one unnecessary request, it is worth it:

- the benefits of having the player loaded on the server side first and then in the server side rendered page are worth it
- searching the name for a uuid is a very cheap request
- the request is still  cached, as it was performed during the SSR

matthias-luger (changes requested):

components/PlayerDetailsList/PlayerDetailsList.tsx
ThomasWu931: Ah, understood. In that case, would it be a better idea to also add a `bids` field down the line as well? So that we can render both the initial 12 bids and auctions on the server side

matthias-luger: Possible, but not necessary, as the initially rendered page is always the auction page.
So the bids dont get rendered, even if they are available. => Bids are always loaded, even when not needed.

pages/player/[uuid]/index.tsx
Comment (ThomasWu931) [reply pages/player/[uuid]/index.tsx:51]: I'm having some trouble reproducing this bug on the localhost development server.
The steps I took was searching for `http://localhost:3000/player/00462df2aabf4f9e8673604970dc8195` before swapping to `http://localhost:3000/player/0a683fa6e4ef4cd1bc0d0754ffeccaa9`. 

From the results that I observed, props.player.uuid had updated. As a result, useEffect was properly called in index.tsx and the selectedPlayer was properly updated.

components/PlayerDetailsList/PlayerDetailsList.tsx
ThomasWu931: This a mistake on my end. Yes, this section of code should be placed inside of the `if`

pages/player/[uuid]/index.tsx
Comment (matthias-luger) [reply pages/player/[uuid]/index.tsx:51]: Yeah, seems like it does update the props...
Never mind then

components/PlayerDetailsList/PlayerDetailsList.tsx
ThomasWu931: I might be wrong but I feel that it would be possible that the initially rendered page is bids. This would happen when searching for player X while on player Y's bid's section. This occurs bc prevDetailType is preserved accross player searches. 

image

PR #890, opened 2022-08-09
- prevent flickering by adding height to the images

### LINKED ISSUE #863: Make search feel faster
The search currently feels slow on first use. The results move multiple times when images are loaded.
The search result (words) should stay in the exact same place when images are not loaded, are loaded or fail to load. 
To optimize further the search image may be a (gray) placeholder until the results are retrieved and displayed

matthias-luger:
fixes #863


Ekwav (approved):
LGTM! 👍

PR #892, opened 2022-08-10
subscribe coflcoin changes

matthias-luger:


matthias-luger:
@Ekwav The websocket doesnt receive a message after the coflcoins change

Flou21:
is this still relevant?

PR #899, opened 2022-08-25
change login to react-oauth/google

matthias-luger:


Ekwav (changes requested):
Requests a relogin on every navigation

Ekwav (changes requested):
The login in the top right pops up on site load instead of when an user actually would need to login.  
It also can't be reopened if the X is clicked once.
Can the login select only be triggered if the user actually wants/needs to login, or at least retrigger after clicking it away once.
I also needed to re-login after returning from standby

matthias-luger:
1. The top right thing only comes up if you have already been logged in (one tap login)
2. If the X is clicked you are not getting logged in, therefore you have the normal login button like you have never been logged in
3. I dont know if I understand this point correctly, but that should be how it works?
4. I dont know what you really mean by standby? Not doing anything over an hour?

Please lets talk about the things in person

Ekwav (approved):

PR #902, opened 2022-08-31
set cache headers

matthias-luger:


utils/CacheUtils.tsx
```diff
@@ -72,6 +72,10 @@ let cacheUtils: CacheUtils = {
 }
 export default cacheUtils
 
+export function getCacheContolHeader() {
+    return 'public, max-age=300, s-maxage=60, stale-while-revalidate=59'
```
Ekwav:
```suggestion
    return 'public, max-age=60, s-maxage=20, stale-while-revalidate=19'
```
we update data more frequently

Ekwav (changes requested):

PR #919, opened 2022-09-18
stop feedback from submit if nothing was changed

### LINKED ISSUE #917: only send valid feedback
only send feedback when at least one change was made
changed a checkbox, inserted text, etc.

matthias-luger:
fixes #917


Ekwav:
The issue asks for the feedback to be submitted if modified, not for the button to block. 
Is the submit button blocked intentionally or did you miss understand the issue?
The current implantation could lead users to change something just for the button to be enabled and thus the feedback would still be useless.

Ekwav (commented):
The issue asks for the feedback to be submitted if modified, not for the button to block. 
Is the submit button blocked intentionally or did you miss understand the issue?
The current implantation could lead users to change something just for the button to be enabled and thus the feedback would still be useless.

PR #922, opened 2022-09-19
show all premium prodcuts on account and premium page

### LINKED ISSUE #918: Display expiry time of lower tier if it differs from current one
`/premium` should display the expiry time of the lower tier if it differs from the current tier to give our users the piece of mind that they didn't loose their time.

matthias-luger:
fixes #918


Ekwav (approved):
LGTM! :+1:

PR #939, opened 2022-09-30
add button to refresh blacklist item names

### LINKED ISSUE #923: Add button to refresh blacklist titles
Items like the one introduced with the nether update in April started out saying `not on ah` but they are actionable and would have a different title now.

Image

there should be a button like the `Clear list` to refresh all entries on the blacklist for new titles.

matthias-luger:
fixes #923


Ekwav (approved):
LGTM! :)

PR #940, opened 2022-09-30
resolve equal api requests with the same response instead of rejecting

### LINKED ISSUE #938: Auction details not rendered in flipper
Clicking on a flip on `/flipper` doesn't work

matthias-luger:
fixes #938


Ekwav (approved):
Works :)

PR #951, opened 2022-10-08
remember when pressing enter while searching

### LINKED ISSUE #948: Wait for search result to arrive
The first element in the search can be opened by pressing enter.
However this currently doesn't wait for the results to arrive leading to either no search result or an incomplete to be opened.
The ender should wait for the last request to complete and only then redirect to the top result

matthias-luger:
fixes #948

### DIFF
--- a/components/Search/Search.tsx
+++ b/components/Search/Search.tsx
@@ -96,9 +96,8 @@ function Search(props: Props) {
                 setResults(searchResults)
                 setIsSearching(false)
 
-                // TODO: Fix with slow connection and test for race condition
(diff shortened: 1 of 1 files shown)


Ekwav (changes requested):
Doesn't redirect when request is loaded but blocks any enter presses until it is.  
It should "save" the enter and redirect when the request completed

Ekwav:
If there are no search results the implementation redirects to the the next search result it finds.
This may cause issues in the future but it's a negligible side effect for now.

PR #963, opened 2022-10-25
let premium users see more recent/active auctions

### LINKED ISSUE #961: Allow paying users to display more auctions
image
Users with starter premium or higher should get more than the default 12 results.  
The two api endpoints for recent and active auctions support pagination of up to 10 pages.  
Users with starter premium should get up to 60 results and users with premium or higher 120.  
There should be a note for users without premium that they can get more results by supporting the project with starter premium.  (Equivalent message for starter premium users scrolling all the way down to the end of their 60)   
Also add the feature to the /premium comparison table

matthias-luger:
fixes #961


Ekwav (changes requested):
* missing loading more active auctions 
* requires the login without the login promt being visible hiding any toasts 
image

* Doesn't display info that more entries require starter plan when logged in
* Same for starter premium

components/ActiveAuctions/ActiveAuctions.tsx
```diff
@@ -60,7 +99,51 @@ function ActiveAuctions(props: Props) {
...
+        })
+    }
+
+    function getMoreActiveAuctionsElement() {
+        if (!isLoggedIn || !premiumType) {
+            return (
+                <p>
+                    You can see more auctions with{' '}
+                    <Link href={'/premium'} style={{ marginBottom: '15px' }}>
+                        <a>Premium:</a>
```
Ekwav: please make this appear as link and remove the `:` currently it looks a bit confusing
image
It looks correct on active tab but not on recent

components/ActiveAuctions/ActiveAuctions.tsx
```diff
@@ -60,7 +99,51 @@ function ActiveAuctions(props: Props) {
...
+                </p>
+            )
+        }
+        if (premiumType.priority === PREMIUM_RANK.STARTER) {
+            return (
+                <p>
+                    You currently use Starter Premium. You can see up to 120 active auctions with
+                    <Link href={'/premium'}>
+                        <a>Premium:</a>
+                    </Link>
```
Ekwav: Same as above except here a space is missing as well
image

components/RecentAuctions/RecentAuctions.tsx
```diff
@@ -81,6 +116,50 @@ function RecentAuctions(props: Props) {
...
+                    You currently use Starter Premium. You can see up to 120 active auctions with
+                    <Link href={'/premium'}>
+                        <a>Premium:</a>
+                    </Link>
+                    <div style={{ marginTop: '15px' }}>
+                        <GoogleSignIn onAfterLogin={onAfterLogin} />
+                    </div>
+                </p>
+            )
+        }
```
Ekwav: This is fully redundant with the active auctions function. Please make it less redundant. (you also didn't change `active auctions` in line 151)

Ekwav (changes requested):

Ekwav (approved):
Still don't like the dupplication of eg. the load more button, there are users waiting for this so I am gonna compromise on that

PR #965, opened 2022-10-27
implemented a filter for auctions and bids lists

### LINKED ISSUE #960: Make player history filterable
The bids and auctions endpoints support all available filters.  
It should be possible for players to filter their own auctions & bids and any player if they have any paid plan (starter-premium and up)   
The UI should look similar to the one for blacklists

matthias-luger:
fixes #960


Ekwav (changes requested):
* Only own auction should be filterable if no paid account (and note of this fact) is missing
* Wiki and hyauctions link should not be displayed 
image
* I am not quite sure but when adding and removing filters the filtered results stay and aren't refreshed currently

pages/player/[uuid]/index.tsx
```diff
@@ -160,6 +156,23 @@ function PlayerDetails(props: Props) {
...
+                        hoverPlacement="bottom"
+                        type="hover"
+                        tooltipContent={
+                            <>
+                                <p>Claim your account to filter your own auctions/bids.</p>
+                                <p>If you have starter premium or above you are able to use the filter for any player.</p>
+                            </>
+                        }
+                    />
+                ) : null}
```
Comment (Ekwav) [inline pages/player/[uuid]/index.tsx]: please move this down below the buttons (where the filter would be) so users see it where they would search for the filter option

Ekwav (changes requested):

components/PlayerDetailsList/PlayerDetailsList.tsx
```diff
@@ -316,6 +317,23 @@ function PlayerDetailsList(props: Props) {
...
+                        </p>
+                    }
+                    hoverPlacement="bottom"
+                    type="hover"
+                    tooltipContent={
+                        <>
+                            <p>Claim your account to filter your own auctions/bids.</p>
+                            <p>If you have starter premium or above you are able to use the filter for any player.</p>
+                        </>
+                    }
```
Ekwav: bad centered
should be below the information icon not in the center.  
Maybe add a redirection possibility to /premium as well

Ekwav (changes requested):

Ekwav (approved):

PR #970, opened 2022-11-04
use itemName property

### LINKED ISSUE #966: Use itemName on /crafts
Currently the tag (internal hypixel name) is used to infer the item name. The `itemName` property should be used instead.
Note: It is prefixed with the color code of the item.

matthias-luger:
fixes #966


Ekwav (approved):
LGTM! :)

PR #975, opened 2022-11-14
added user flip finder

### LINKED ISSUE #974: Add user flip finder
The `USER` finder (value `16`) should be added to the selectable finders with the following description:
```
Forwards all new auctions with a target value set to the starting bid (0 profit)
You can use this together with whitelist/blacklist of <b>Starting Bid</b> and other filters to create your own flip rules.
Different to the other finders this one won't pre-filter auctions its all up to you.
```

matthias-luger:
fixes #974


Ekwav (approved):
LGTM! :)

PR #980, opened 2022-11-18
Censor kat flips

matthias-luger:
To decrease reports regarding the kat flips, I censored the data UI-side. The amount of people checking the request should be neglectable.


Ekwav (changes requested):
I don't see an issue in once report a quarter.  
If you want to hide it completely please include a note that the top 3 can be unlocked with starter premium which costs just 1.20 a month

Ekwav (approved):

PR #984, opened 2022-11-24
black friday purchase ui

matthias-luger:


components/CoflCoins/CoflCoinsPurchase.tsx
```diff
@@ -65,7 +65,14 @@ function Payment(props: Props) {
         />
     )
```
Ekwav, line 67: Please add a countdown for how long the prices last to expire at 2022-11-26 06:00 UTC

Ekwav (changes requested):

PR #995, opened 2022-11-29
Faster premium product load

### LINKED ISSUE #991: Speed up rendering of the filter
The item filter on `/player/{uuid}` should load faster
image
Currently it takes more than two seconds.
image

I suggest caching the expiry of premium locally and refreshing it asynchronously

matthias-luger:
fixes #991


Ekwav (approved):
LGTM! :)

PR #1004, opened 2022-12-14
remove flip filters from player auction/bids

### LINKED ISSUE #1002: Don't show flipper filters on player page
Image

There is no profit on bids thus the backend returns an error.  
That and the other flip filters should not be shown on the player page

matthias-luger:
fixes #1002


Ekwav (approved):
LGTM! :)

PR #1020, opened 2023-01-03
hide sold preloaded flips

### LINKED ISSUE #1019: Make sure sells are hidden if setting is on
Image

matthias-luger:
fixes #1019

### DIFF
--- a/components/Flipper/Flipper.tsx
+++ b/components/Flipper/Flipper.tsx
@@ -44,9 +44,15 @@ interface Props {
 }
 
 function Flipper(props: Props) {
-    let [flips, setFlips] = useState<FlipAuction[]>(props.flips || [])
-    let [isLoggedIn, setIsLoggedIn] = useState(false)
     let [flipperFilter, setFlipperFilter] = useState<FlipperFilter>(getSettingsObject<FlipperFilter>(FLIPPER_FILTER_KEY, {}))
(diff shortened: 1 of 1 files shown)


components/Flipper/Flipper.tsx
```diff
@@ -44,9 +44,15 @@ interface Props {
 }
 
 function Flipper(props: Props) {
-    let [flips, setFlips] = useState<FlipAuction[]>(props.flips || [])
-    let [isLoggedIn, setIsLoggedIn] = useState(false)
     let [flipperFilter, setFlipperFilter] = useState<FlipperFilter>(getSettingsObject<FlipperFilter>(FLIPPER_FILTER_KEY, {}))
+    let [flips, setFlips] = useState<FlipAuction[]>(
```
Ekwav, line 48: What about new flips getting sold?
The statehook is only created once when the site loads, isn't it?

Ekwav (changes requested):

components/Flipper/Flipper.tsx
matthias-luger: If a new flip comes in, it is checked for being sold: 
grafik

Also, if I get the websocket event that a flip was sold, it is removed, if the checkbox is checked: 
grafik

Ekwav (approved):
LGTM! :)

PR #1026, opened 2023-01-10
add special case for PET_SKIN

### LINKED ISSUE #1023: Don't change Pet skin category
This filter value should not be changed, it causes massive confusion.
It should be displayed as `Pet Skin (unapplied)`
To highlight it's unapplied pet skins.
https://github.com/Coflnet/SkyItems/blob/main/Models/ItemCategory.cs#L66

matthias-luger:
fixes #1023


Ekwav (approved):
LGTM! 👍

PR #1030, opened 2023-01-11
add rate limiting

matthias-luger:


pages/_middleware.tsx
```diff
@@ -0,0 +1,22 @@
...
+import requestIp from 'request-ip'
+
+const limiter = rateLimit({
+    interval: 60 * 1000, // 60 seconds
+    uniqueTokenPerInterval: 500 // Max 500 users per second
+})
+
+export async function middleware(req: NextRequest, ev: NextFetchEvent) {
+    try {
+        const clientIp = requestIp.getClientIp(req)
```
Ekwav, line 12: Aren't you reading the cloudflare with this?
There is a header that cloudflare uses to forward the real client IP address

matthias-luger: @Ekwav The package should handle that, shouldnt it? https://www.npmjs.com/package/request-ip
grafik

Ekwav (approved):

PR #1036, opened 2023-01-17
add prop to disable localstorage use for itemfilter

### LINKED ISSUE #1028: Don't copy and auto apply filter from blacklists on user page
Currently the last added blacklist filter rule configured on `/flipper` is automatically copied and applied on `/player/{uuid}`.  
This doesn't fit how our users use the site blacklists are different from past auctions.   
Filters from the flipper rules configuration should not be carried over automatically.  
They should not be applied at all or only be copied/applied when the user chooses to do so via a button or similar.

matthias-luger:
fixes #1028


Ekwav (approved):
Works :)

PR #1041, opened 2023-01-24
tracked flips are now formatted using the color codes in the name

### LINKED ISSUE #1040: Fix best flip preview
https://sky.coflnet.com/player/df2b974944b7416899a1ec1fe396519a/flips
There are two issues:
* The best flip is not the best flip (not sorted)
* The item name still contains formatting colors which should be removed

matthias-luger:
fixes #1040


Ekwav (approved):
LGTM but checks fail

PR #1042, opened 2023-01-24
searchable select input filters are now not prefilled

### LINKED ISSUE #1038: Make it easier to change selection
Currently multi option drop-down search selects are pre filled with the first option.
While that is great if the intention is to select the first option it very much hinders the selection of another option because the prefilled one is text and has to be deleted manually.
It should be possible to change the selected option with fewer actions either by not actually filling the first option in, by ignoring the current input in the search unless changed (showing all) or by deleting the input on focus

matthias-luger:
fixes #1038

### DIFF
--- a/components/FilterElement/FilterElement.tsx
+++ b/components/FilterElement/FilterElement.tsx
@@ -152,7 +152,7 @@ function FilterElement(props: Props) {
                     />
                 )
             } else {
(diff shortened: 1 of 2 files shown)


Ekwav (approved):
LGTM! 🙃

PR #1043, opened 2023-01-24
add profit percent to tracked flips

### LINKED ISSUE #1037: Add profit percentage sort option
There should be a profit percent sorting option on `/player/{uuid}/flips`
image

matthias-luger:
fixes #1037


Ekwav (commented):
👍

PR #1063, opened 2023-02-15
removed default React import

ashishsurya:
Opened a new pull request after merging of #1047 @matthias-luger


matthias-luger (changes requested):
The build fails because the "React"-Import was removed in a file where it was needed.
Specifically in "FlipCustomize": `export default React.memo(FlipCustomize)`

matthias-luger:
It may be needed in even more files. Just execute `npm install` and then `npm run build` to see if the build runs through

ashishsurya:
@matthias-luger done 
commit ->  75f139756a4b574a9d206054061420eb793345be

PR #1073, opened 2023-03-05
Add no adjust to purse

### LINKED ISSUE #1071: Add NoAdjustToPurse setting
The setting with key `modnoAdjustToPurse` should be added to the settings to be select-able by the user.  
It disables the default adjustment to purse of maxcost.

matthias-luger:
fixes #1071

### DIFF
--- a/components/Flipper/FlipCustomize/FlipCustomize.tsx
+++ b/components/Flipper/FlipCustomize/FlipCustomize.tsx
@@ -516,6 +516,21 @@ function FlipCustomize() {
                                     type="checkbox"
                                 />
                             </Form.Group>
+                            <Form.Group>
+                                <Form.Label className={styles.label} htmlFor="modCountdown">
+                                    Automatic maxCost
+                                </Form.Label>
+                                <Form.Check
+                                    onChange={event => {
+                                        updateApiSetting('modNoAdjustToPurse', !event.target.checked)
+                                        setFlipCustomizeSetting('modNoAdjustToPurse', !event.target.checked)
+                                    }}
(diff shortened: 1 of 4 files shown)


components/Flipper/FlipCustomize/FlipCustomize.tsx
```diff
@@ -516,6 +516,21 @@ function FlipCustomize() {
                                     type="checkbox"
                                 />
                             </Form.Group>
+                            <Form.Group>
+                                <Form.Label className={styles.label} htmlFor="modCountdown">
```
Ekwav: If I am not mistaken, that htmlFor targets the wrong input

components/Flipper/FlipCustomize/FlipCustomize.tsx
```diff
@@ -516,6 +516,21 @@ function FlipCustomize() {
...
+                                    Set maxCost to purse
```
Ekwav:
```suggestion
                                    Don't show flips costing more than purse
```
The behavior will probably change to the purse not modifying maxCost because many settings updates cause issues

Ekwav (changes requested):

components/Flipper/FlipCustomize/FlipCustomize.tsx
matthias-luger: I changed it to `Don't show flips that <br /> cost more than purse` because the width of the text doesnt fit well into the limited space of the labels

Ekwav (approved):
LGTM 👍

PR #1075, opened 2023-03-08
only store name and tag in settings

### LINKED ISSUE #1065: Optimize settings size
The `iconUrl` on black and whitelists could be removed as it can be calculated from the `tag` prefixed with `https://sky.coflnet.com/static/icon` to make settings smaller.
I could not find a reason for the iconUrl to be included, please let me know if I missed it.

matthias-luger:
fixes #1065


Ekwav (approved):
LGTM! :)

PR #1078, opened 2023-03-08
accept server side settings for import

### LINKED ISSUE #1066: Support importing settings in backend format
To import flips from server side backups and allow others to modify one coherent json object the format sent over the socket should be supported in imports (and if possible also be the export format from now on)

matthias-luger:
fixes #1066


matthias-luger:
I tested it with the attached file (.txt so I can attach it here, normally .json).
I currently check for the field "whitelist" to be a property on the parsed object to detect if it is in the serverside format. I expect that it is always there even if empty, right? Otherwise tell me and I will use another property


[serverside-filter.txt](https://github.com/Coflnet/hypixel-react/files/10925840/serverside-filter.txt)

Ekwav (changes requested):
The server format always has the whitelist property, it may be `null` but it exists.
What missing is that the export should use the same format aka one object to allow easier editing by users

Ekwav (approved):
LGTM!

PR #1079, opened 2023-03-08
fix resetting list after order change

### LINKED ISSUE #1064: Sort order does not change
image
this select does not change the order, there no request triggered

matthias-luger:
fixes #1064


Ekwav (approved):
LGTM! :)

PR #1084, opened 2023-03-27
add sorting to restrictions

matthias-luger:


components/Flipper/FlipRestrictionList/FlipRestrictionList.module.css
```diff
@@ -36,6 +36,21 @@
...
+    position: relative !important;
+}
+
+.sortByNameContainer{
+    display: inline-flex;
+    margin-left: 20px;
+}
+
+.searchFilter{
+    width: 49%;
```
Ekwav: why is this 49%?
I assume you want to make it half of the container?
How about flexbox/css-gird?

Ekwav (changes requested):

Ekwav (approved):
:+1:

PR #1088, opened 2023-03-30
Next 13 upgrade

matthias-luger:
update to next 13
update basically all other npm packages as well


matthias-luger:
@Ekwav Never mind the review. While it seems like the production build runs fine (at least it looks like it), the dev build still throws a lot of thing that I still need to fix. 
So the review sadly has to wait for now

matthias-luger:
@Ekwav I think this is ready for you to take a look now if you want :)

Ekwav (changes requested):
The tests should pass

matthias-luger:
@Ekwav Took a few runs but the tests ran through

PR #1170, opened 2023-10-19
Add lemonsqueezy payment provider

### LINKED ISSUE #1169: Add Lemonsqueezy payment provider
On /premium there should be a dropdown to select the current country (default to what language is set) 
If the selected country is outside of the EU/US/UK the options should be only lemonsqueezy. 
For countries inside that group the current option should stay the same.
Lemonsqueezy prices and products are the same as paypal except that they start with an `l` instead of a `p`

matthias-luger:
fixes #1169

### DIFF
--- a/api/ApiHelper.tsx
+++ b/api/ApiHelper.tsx
@@ -1015,6 +1015,44 @@ export function initAPI(returnSSRResponse: boolean = false): API {
         })
     }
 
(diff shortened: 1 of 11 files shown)


Ekwav (commented):
LGTM code wise, could not build it.

> npm ERR! gyp ERR! UNCAUGHT EXCEPTION 
> npm ERR! gyp ERR! stack Error: Cannot find module '@gar/promisify'

Issue #1114, opened 2023-05-30
Add option to only show same buyer


Ekwav:
The flips page should have an option to toggle to only showing flips where the buyer matches the seller (seller is the identifier in the url as well)

matthias-luger:
@Ekwav I am sorry, I have trouble understanding this feature?

This is an example flip from a random user d06eccea361a44188d7d69f80b4b6a37. What exactly should match to what?

```
{
      "itemName": "§dLoving Crimson Chestplate",
      "itemTag": "CRIMSON_CHESTPLATE",
      "tier": "MYTHIC",
      "pricePaid": 7400000,
      "soldFor": 6381408,
      "finder": 0,
      "uId": -482164237665047230,
      "originAuction": "5aa3df989be8413eaf85c40e6fb30559",
      "soldAuction": "f94f022e32934ed2a13ff9783f49d337",
      "buyTime": "2023-10-13T02:13:52",
      "sellTime": "2023-10-15T15:19:24.26",
      "profit": -12541096,
      "propertyChanges": [
        {
          "description": "ah tax",
          "effect": -128828
        },
        {
          "description": "Recombobulator",
          "effect": -7011118
        },
        {
          "description": "Enchant ultimate_wisdom 5",
          "effect": -2500003
        },
        {
          "description": "Reforge Loving added",
          "effect": -1882555
        }
      ]
    }
```

Ekwav:
I wanted to combine this with https://github.com/Coflnet/SkyFlipTracker/issues/25 and add a field/list of flags of properties (is same buyer, found price from trade etc)
So this depends on that

PR #1191, opened 2023-11-29
add flag filter to flip tracking

### LINKED ISSUE #1188: Add option to only show trade/non trades
The flips page should have an option to toggle to only showing flips where the `ViaTrade` Flag is set.   
Related to https://github.com/Coflnet/hypixel-react/issues/1114 both flags now exist.  
There also is a third flag indicating if only one item or multiple were traded (price fraction is estimated in that case)

matthias-luger:
fixes #1188


matthias-luger:
fixes #1114
fixes #1188

PR #1192, opened 2023-11-30
Importing BM Config

jason-tung:
Add config importing for BM Config

Works and only some small edge cases are broken but I'd rather fix those on a by-report basis because there are too many misc items and usually it's the fault of the filter creator.


jason-tung:
[sam copy.json](https://github.com/Coflnet/hypixel-react/files/13551849/sam.copy.json)
a popular filter to test with

PR #1195, opened 2023-12-09
BM Filter - Add Support for Runes

jason-tung:
Added support for most runes excluding a few that still need a level attribute. Already raised issue with ekwav.

image


matthias-luger:
LGTM!

PR #1202, opened 2024-01-01
Delay to minProfit

Conutik:
Added a delay to changing the minProfit value


matthias-luger:
I fixed it with 4d3dc70786b88fe5563c7384063b4e3970f39afd
The change will be live next time I roll out the current version (most likely tomorrow or so)
I didn't use your Pull Request as I didn't want to add a external package for something little like this :)

PR #1278, opened 2024-06-28
added craftcost finder

### LINKED ISSUE #1272: Add CraftCost finder to options
There is a new finder called `CraftCost` its description is:
```
Displays any auction that would be at least 20% more expensive to craft.
Sums up clean+modifier cost.
Note that this does not indicate that the item will sell for that price.
```

matthias-luger:
fixes #1272

### DIFF
--- a/utils/FlipUtils.tsx
+++ b/utils/FlipUtils.tsx
@@ -148,7 +148,21 @@ export const FLIP_FINDERS = [
         ),
         selectable: true
     },
-    { value: '128', label: 'External', shortLabel: 'External', default: false, description: '', selectable: false }
+    { value: '128', label: 'External', shortLabel: 'External', default: false, description: '', selectable: false },
+    {
+        value: '1024',
+        label: 'CraftCost',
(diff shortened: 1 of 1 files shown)


utils/FlipUtils.tsx
```diff
@@ -148,7 +148,21 @@ export const FLIP_FINDERS = [
...
-    { value: '128', label: 'External', shortLabel: 'External', default: false, description: '', selectable: false }
+    { value: '128', label: 'External', shortLabel: 'External', default: false, description: '', selectable: false },
+    {
+        value: '1024',
+        label: 'CraftCost',
+        shortLabel: 'CraftCost',
+        default: false,
+        description: (
+            <span>
+                Displays any auction that would be at least 20% more expensive to craft. <br />
```
Ekwav:
```suggestion
                Displays any auction that would be at least 20% or 10m more expensive to craft. <br />
```

Ekwav (changes requested):
The definition already changed :D

PR #1280, opened 2024-07-01
reiterate->duplex.tsx

Parkusisafk:


components/AuctionDetails/AuctionDetails.tsx
```diff
@@ -457,6 +457,8 @@ function AuctionDetails(props: Props) {
                         <ul className={styles.list}>
                             {auctionDetails?.enchantments.map(enchantment => {
                                 let enchantmentString = <span>{enchantment.name}</span>
+                                enchantment.name=enchantment.name.replace("Ultimate_Reiterate","Duplex");
+                                enchantment.name=enchantment.name.replace("Ultimate Reiterate", "Duplex");
```
matthias-luger:
```suggestion
                                if(enchantment.name === "Ultimate Reiterate"){
                                    enchantment.name = "Ultimate Duplex";
                                }
```

matthias-luger (changes requested):
1. Don't use the replace function on all enchantments, as most of the time it is not necessry. A simple check for the name should be enough (see my suggestion)
2. Why do you check for "Ultimate_Reiterate" as well? As far as I know it should always be without underscore
3. Put the replace above line 459. In practice it should not matter as there is always a color, but the element above would be wrong, it there wouldn't be a color present

matthias-luger:
@Parkusisafk friendly reminder

components/AuctionDetails/AuctionDetails.tsx
Parkusisafk: got it!

Parkusisafk: I checked the hypixel api and the raw output I'm getting is "Ultimate_Reiterate"
All good though the underscore should have been removed while indexing I guess

Parkusisafk:
got it!


I checked the hypixel api and the raw output I'm getting is "Ultimate_Reiterate"
All good though the underscore should have been removed while indexing I guess

Parkusisafk:
that was my bad I didn't read the code thoroughly

components/AuctionDetails/AuctionDetails.tsx
```diff
@@ -456,7 +456,11 @@ function AuctionDetails(props: Props) {
                     {auctionDetails && auctionDetails!.enchantments.length > 0 ? (
                         <ul className={styles.list}>
                             {auctionDetails?.enchantments.map(enchantment => {
+                                if(enchantment.name === "Ultimate Reiterate"){
+                                    enchantment.name = "Duplex";
```
matthias-luger: Is there a specific reason you just called it Duplex? We show all the Ultimate enchants with the "Ultimate" prefix to make it more obvious and easier to find (for example in the item filter search)

matthias-luger: @Parkusisafk After this is cleared, I will merge the PR :)

Parkusisafk: oops, did ya'll recently changed ultimate enchants to add the prefix "Ultimate"? When I was doing the change I remember that I stumbled across a terminator with just "Soul Eater", so I assumed that the ultimate enchants did not have the "Ultimate" prefix, and instead "Ultimate Reiterate" is just some raw data carried over from the api with the "Ultimate" still on it. Just now I checked and indeed the prefix is present. Thank you for correcting me

To address the ".replace" function, at first I was unsure of the content of the string [enchantment.name], ie. there might be other characters inside (example: Enchantment_Ultimate_Reiterate), and there might be other functions such as .slice that removes the extra part and extract the enchantment. So putting ".replace" was just being cautious

matthias-luger:
LGTM!

PR #1326, opened 2024-11-02
Order premium types highlight time difference and hide irrelevant tiers

### LINKED ISSUE #1325: change premium time display to be additive
don't display that premium time is 10 days if there is 7 days of prem+ instead state that after prem+ there is 3 days of premium, also don't display a lower tier if there is less than 1% time difference
context:https://discord.com/channels/267680588666896385/1301724455533547542/1301884857428869133

Ekwav:
fix #1325

### DIFF
--- a/utils/PremiumTypeUtils.tsx
+++ b/utils/PremiumTypeUtils.tsx
@@ -8,6 +8,13 @@ export const TEST_PREMIUM_DAYS = 3
 
 export const PREMIUM_TYPES: PremiumType[] = [
     {
+        productId: 'premium',
+        label: 'Premium',
+        durationString: 'month',
+        priority: PREMIUM_RANK.PREMIUM,
+        options: generateNumberOptionArray(1, 12, 'premium', 1800)
+    },
+    {
         productId: 'premium_plus',
         label: 'Premium+',
         durationString: '',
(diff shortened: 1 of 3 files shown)


utils/PremiumTypeUtils.tsx
```diff
@@ -7,13 +7,6 @@ export enum PREMIUM_RANK {
 export const TEST_PREMIUM_DAYS = 3
 
 export const PREMIUM_TYPES: PremiumType[] = [
-    {
-        productId: 'premium',
-        label: 'Premium',
-        durationString: 'month',
-        priority: PREMIUM_RANK.PREMIUM,
-        options: generateNumberOptionArray(1, 12, 'premium', 1800)
-    },
```
matthias-luger: This might break the current transaction history logic in TransactionHistory.tsx:
If the transaction productId is "premium_plus....", it first matches with premium_plus and then with premium. Until now that wasn't a problem as premium was the first to be compared

```
 let purchasedPremiumOption
        PREMIUM_TYPES.forEach(premiumType => {
            if (transaction.productId.startsWith(premiumType.productId)) {
                purchasedPremiumOption = premiumType.label
            }
        })
```

matthias-luger (changes requested):

utils/PremiumTypeUtils.tsx
Ekwav: Ah, I see, good catch. I reverted my changes.

PR #1392, opened 2025-07-29
Merge flips into flipping hub

Ekwav:


matthias-luger:
@Ekwav Do you really want to make it that difficult to navigate to these flip pages? You know need to A) Know that the pages are "hidden" in that hub and you always need 2 clicks?

What do you think about a sub-navigation, so when hovering over "flipping-hub" i can directly navigate to these pages?

Ekwav:
> @Ekwav Do you really want to make it that difficult to navigate to these flip pages? You know need to A) Know that the pages are "hidden" in that hub and you always need 2 clicks?
> 
> What do you think about a sub-navigation, so when hovering over "flipping-hub" i can directly navigate to these pages?

I don't know how to do that, you can add that if you want @matthias-luger

Issue #1397, opened 2025-08-11
Add bazaar flip page


Ekwav:
/api/flip/bazaar/spread should be displayed on a new bazaar flip page it contains the current estimated profit profit on a weekly basis.

/api/flip/bazaar/spread/deemand is current demand based and should be displayed to premium users

Ekwav:
The current draft in the bazaar-flips branch lags a lot, I don't think it should render outside of the viewport

Ekwav:
@matthias-luger it should render at least some samples server side

<img width="1955" height="1325" alt="Image" src="https://github.com/user-attachments/assets/aa5a9555-01b2-4a86-9d7a-04a5e11fcb5c" />

matthias-luger:
SSR fixed by 5755d674470a5fc90d1204d68c7bd7749d93e581

PR #1413, opened 2025-08-22
Bazaar flips

matthias-luger:
relates to #1397


Ekwav (changes requested):
The page currently lags a lot, especially noticeable when changing the sorting

PR #1430, opened 2025-09-26
setup compliant ad integration

Ekwav:
fix https://github.com/Coflnet/internal/issues/408


hooks/useShouldShowAds.ts
```diff
@@ -0,0 +1,60 @@
...
+
+    useEffect(() => {
+        function handlePremiumUpdate() {
+            if (!isMountedRef.current) {
+                return
+            }
+            setShouldShowAds(evaluateShouldShowAds())
+        }
+
+        window.addEventListener('premium.products.updated', handlePremiumUpdate as EventListener)
```
matthias-luger: I dont like the use of custom events, but i assume it is fine in this case.
But at least store the key as a constant in the CUSTOM_EVENTS

components/Ads/NitroAdSlot.tsx
```diff
@@ -0,0 +1,89 @@
...
+    className?: string
+    style?: CSSProperties
+}
+
+export default function NitroAdSlot({ slotId, config, className, style }: NitroAdSlotProps) {
+    const shouldShowAds = useShouldShowAds()
+    const configRef = useRef(config)
+    const hasRequestedRef = useRef(false)
+
+    useEffect(() => {
```
matthias-luger: this should not be in a useEffect -> see any useRef i have used anywhere

components/AdScriptLoader.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+    return (
+        <>
+            <Script
+                data-cfasync="false"
+                dangerouslySetInnerHTML={{
+                    __html:
+                        "window.nitroAds=window.nitroAds||{createAd:function(){return new Promise(e=>{window.nitroAds.queue.push(['createAd',arguments,e])})},addUserToken:function(){window.nitroAds.queue.push(['addUserToken',arguments])},queue:[]};"
+                }}
+            />
```
matthias-luger, line 26: O.o

matthias-luger (changes requested):

matthias-luger:
I quit the review midway through. 

1) Please remove all the AI comments
2) Please rework the use of custom events -> 1 for some global notification regarding the ads is ok if it is really needed, but otherwise write a provider, or solve it another way
3) rework the use of effects. You are using useEffect way to excessive: https://react.dev/learn/you-might-not-need-an-effect

Ekwav:
> I quit the review midway through.
> 
> 1. Please remove all the AI comments
> 2. Please rework the use of custom events -> 1 for some global notification regarding the ads is ok if it is really needed, but otherwise write a provider, or solve it another way
> 3. rework the use of effects. You are using useEffect way to excessive: https://react.dev/learn/you-might-not-need-an-effect

Did those those I think, can you point to a case thats still wrong?

components/AdScriptLoader.tsx
Ekwav: 🤨

Ekwav:
@matthias-luger I am gonna merge this this weekend, in case you still have change requests please voice them now

PR #1431, opened 2025-09-29
Let copilot come up with a fix for flanky test

Ekwav:


matthias-luger:
this much code that is just a workaround for one test is not acceptable. It basically just adds special cases to mock the entire thing. I will take another look at the crafts tests, but i would rather remove it that this solution

PR #1434, opened 2025-09-30
Add attribute flips page

Ekwav:


api/attributeFlips.ts
matthias-luger, line 1: why does this file exist? use the generated api

components/AttributeFlips/AttributeFlips.tsx
```diff
@@ -0,0 +1,382 @@
...
+            ) : null}
+            {modifiers && modifiers.length > 0 ? (
+                <div className={styles.badgeRow}>
+                    {modifiers.map(modifier => (
+                        <Badge bg="secondary" key={`${modifier.key}-${modifier.value}`}>
+                            {formatModifier(modifier)}
+                        </Badge>
+                    ))}
+                </div>
+            ) : null}
```
matthias-luger: 1) Add spacing between these two, as they is none if both enchantments and modifiers exist
2) Make sure to use the text doesnt overflow if it is too long (existing CSS-Class: "ellipse")

components/AttributeFlips/AttributeFlips.tsx
```diff
@@ -0,0 +1,382 @@
...
+                    </p>
+                    {messageFromApi ? <p className={styles.apiMessage}>{messageFromApi}</p> : null}
+                </Alert>
+                <div className={styles.ctaRow}>
+                    <Link href="/premium" className="disableLinkStyle" rel="nofollow">
+                        <Button variant="primary">See premium plans</Button>
+                    </Link>
+                    <GoogleSignIn onAfterLogin={handleAfterLogin} />
+                </div>
+            </div>
```
matthias-luger: As already said in another PR: These Premium notifiers should be a reusable component

matthias-luger (changes requested):

api/attributeFlips.ts
Ekwav: I guess the generated code got lost in between, removed the file

components/AttributeFlips/AttributeFlips.tsx
Ekwav: I am not sure what exactly is meant I gave your comment to copilot and it did a thing

components/AttributeFlips/AttributeFlips.tsx
Ekwav: Refactored it into a standalone component

app/attributeFlips/page.tsx
```diff
@@ -4,14 +4,14 @@ import { HydrationBoundary, dehydrate } from '@tanstack/react-query'
...
-import { fetchAttributeFlips } from '../../api/attributeFlips'
+import { getGetApiFlipAttributeQueryOptions } from '../../api/_generated/skyApi'
 
 export default async function Page() {
     const queryClient = getQueryClient()
 
     await queryClient.prefetchQuery({
-        queryKey: ['attributeFlips', 'default'],
-        queryFn: () => fetchAttributeFlips()
+        ...getGetApiFlipAttributeQueryOptions(undefined, { fetch: { cache: 'no-store' } }),
```
matthias-luger: why this weird implementation and not like every other? Something like this:
    queryClient.prefetchQuery({
        queryKey: [getGetApiFlipAttributeQueryKey()],
        queryFn: () => getApiFlipAttributeFlips()
    })

matthias-luger (changes requested):

app/attributeFlips/page.tsx
Ekwav: I removed the prefetch, as far as I can tell the SSR does not know about the google token

PR #1435, opened 2025-09-30
Favourite list

Ekwav:
this uses the same endpoit as #1432 so should maybe be merged after


components/Favorites/FavoriteItemsBar.tsx
```diff
@@ -0,0 +1,197 @@
...
+    let [priceMovements, setPriceMovements] = useState<Record<string, ItemPriceMovement>>({})
+    let [isLoading, setIsLoading] = useState(false)
+    let [error, setError] = useState<string | null>(null)
+
+    useEffect(() => {
+        function load() {
+            setFavorites(getFavoriteItems())
+        }
+
+        load()
```
matthias-luger: what is this load doing? It runs initially, but the default value already sets the favorites variable

components/Favorites/FavoriteItemsBar.tsx
```diff
@@ -0,0 +1,197 @@
...
+            if (Array.isArray(detailFavorites)) {
+                setFavorites(detailFavorites)
+            } else {
+                load()
+            }
+        }
+
+        document.addEventListener(CUSTOM_EVENTS.FAVORITES_UPDATED, onFavoritesUpdated)
+        return () => {
+            document.removeEventListener(CUSTOM_EVENTS.FAVORITES_UPDATED, onFavoritesUpdated)
```
matthias-luger: please dont rely on custom events. If two components have to communicate with each other that are on different places in the tree, use a Provider

components/Favorites/FavoriteItemsBar.tsx
```diff
@@ -0,0 +1,197 @@
...
+            intervalId = setInterval(fetchMovements, REFRESH_INTERVAL)
+        }
+
+        return () => {
+            cancelled = true
+            if (intervalId) {
+                clearInterval(intervalId)
+            }
+        }
+    }, [favorites])
```
matthias-luger: why are you not using the generated api?

components/Favorites/FavoriteItemsBar.tsx
```diff
@@ -0,0 +1,197 @@
...
+                            <div className={styles.actions}>
+                                <SubscribeButton
+                                    topic={favorite.tag}
+                                    type={favorite.bazaar ? 'bazaar' : 'item'}
+                                    buttonContent={<span>Notify</span>}
+                                />
+                                <FavoriteToggle item={item} size="small" />
+                            </div>
+                        </article>
+                    )
```
matthias-luger: This should probably be a own component but that is only my preference

components/Favorites/FavoriteToggle.tsx
```diff
@@ -0,0 +1,83 @@
...
+    useEffect(() => {
+        function handleFavoritesUpdate(_event?: Event) {
+            if (!item?.tag) {
+                return
+            }
+            setIsFavorite(isItemFavorite(item.tag))
+        }
+        document.addEventListener(CUSTOM_EVENTS.FAVORITES_UPDATED, handleFavoritesUpdate)
+        return () => {
+            document.removeEventListener(CUSTOM_EVENTS.FAVORITES_UPDATED, handleFavoritesUpdate)
```
matthias-luger: dont rely on CUSTOM_EVENTS.

Also: Is it possible that the only way this is triggered, is through the "toggleFavoriteItem" function? The function is only used in this component, therefore this component is the only one that can even trigger this event. Why would you need to subscribe to it?

utils/FavoriteItemUtils.tsx
```diff
@@ -0,0 +1,77 @@
...
+export function removeFavoriteItem(tag: string): FavoriteItemEntry[] {
+	if (!isClientSideRendering()) {
+		return []
+	}
+
+	const favorites = getFavoriteItems()
+	const newFavorites = favorites.filter(f => f.tag !== tag)
+	persistFavorites(newFavorites)
+	return newFavorites
+}
```
matthias-luger: these are never used

api/ApiHelper.tsx
```diff
@@ -1667,6 +1668,43 @@ export function initAPI(returnSSRResponse: boolean = false): API {
...
+                    resolve(parsed)
+                },
+                reject: function (error) {
+                    apiErrorHandler(RequestType.PRICE_CHANGE, error, { itemTags })
+                    reject(error)
+                }
+            })
+        })
+    }
+
```
matthias-luger: why not use the generated api  "useGetApiPricesChange"?

components/Favorites/FavoriteItemsBar.tsx
```diff
@@ -0,0 +1,197 @@
...
+import { convertTagToName, numberWithThousandsSeparators } from '../../utils/Formatter'
+import { getFavoriteItems } from '../../utils/FavoriteItemUtils'
+import FavoriteToggle from './FavoriteToggle'
+import SubscribeButton from '../SubscribeButton/SubscribeButton'
+import styles from './FavoriteItemsBar.module.css'
+
+const REFRESH_INTERVAL = 5 * 60 * 1000
+
+function FavoriteItemsBar() {
+    let [favorites, setFavorites] = useState<FavoriteItemEntry[]>(() => getFavoriteItems())
```
matthias-luger: useState<FavoriteItemEntry[]>(getFavoriteItems) would also work the same way

matthias-luger (changes requested):

components/Favorites/FavoriteItemsBar.tsx
Ekwav: didn't tell copilot to do, but now did

components/Favorites/FavoriteItemsBar.tsx
Ekwav: refactored

components/Favorites/FavoriteToggle.tsx
```diff
@@ -25,18 +26,9 @@ function FavoriteToggle({ item, className, showLabel = false, size = 'medium' }:
...
-            setIsFavorite(isItemFavorite(item.tag))
-        }
-        document.addEventListener(CUSTOM_EVENTS.FAVORITES_UPDATED, handleFavoritesUpdate)
-        return () => {
-            document.removeEventListener(CUSTOM_EVENTS.FAVORITES_UPDATED, handleFavoritesUpdate)
-        }
-    }, [item?.tag])
+    // When favorites change elsewhere, the provider will update storage and state; we
+    // reflect that via the shared utils read above. The provider will be responsible
+    // for keeping favorites in sync across the app.
```
matthias-luger: remove AI comments (counts for the entire review)

utils/FavoriteItemUtils.tsx
```diff
@@ -27,6 +25,11 @@ export function isItemFavorite(tag: string | undefined): boolean {
 	return getFavoriteItems().some(f => f.tag === tag)
 }
 
+/**
```
matthias-luger: replace the usage instead of marking it depricate

components/Favorites/FavoriteItemsBar.tsx
```diff
@@ -57,9 +39,14 @@ function FavoriteItemsBar() {
             setIsLoading(true)
             setError(null)
             try {
-                let data = await api.getPriceMovements(favorites.map(f => f.tag))
+                const resp = await getApiPricesChange({ itemTags: favorites.map(f => f.tag) })
```
matthias-luger: like in other PRs, try to use the react-query hooks instead of calling the functions inside a useEffect-hook

matthias-luger (changes requested):

PR #1438, opened 2025-10-04
Add mayor flips

### LINKED ISSUE #1427: Add mayor diff flips
`/api/flip/mayor` contains a list of the expected change of item values after/until the next mayor takes office. It should be possible for users to see those flips estimations

Ekwav:
fixes #1427


app/mayor/page.tsx
```diff
@@ -0,0 +1,46 @@
...
+    try {
+        await queryClient.prefetchQuery({
+            queryKey: [getGetApiFlipMayorQueryKey()],
+            queryFn: () => getApiFlipMayor()
+        })
+    } catch (e) {
+        // If the endpoint requires authentication and we don't have a server-side
+        // token, prefetch can fail with 401/403. Don't block rendering — the
+        // client will re-fetch with the user's token when available.
+        console.warn('Prefetching mayor flips failed (likely unauthenticated):', e)
```
matthias-luger: what do you mean "if the endpoint requires authentication"? As far as i know it does need it, therefore this will always fail and doesnt make sense to prefetch.

components/MayorFlips/MayorFlips.tsx
```diff
@@ -0,0 +1,265 @@
...
+
+    return (
+        <div>
+            <p>
+                Capitalize on Hypixel SkyBlock mayor election cycles with data-driven flips. Our analysis tracks historic price patterns from previous mayor
+                terms to predict which items will surge or drop when the next mayor is elected. Perfect for players looking to maximize profits around election
+                events.
+            </p>
+            {!hasPremium ? (
+                // TODO: combine this with the other premium notices once merged
```
matthias-luger: why add a todo instead of doing it? Or at least crate a issue, because otherwise this will never be done

hooks/useSortedAndFilteredItems.tsx
```diff
@@ -80,7 +81,11 @@ export function useSortedAndFilteredItems<T>(
             try {
                 setIsProcessing(true)
 
-                if (currentItems.length === 0) {
+                // Normalize incoming items to an array to avoid runtime errors when
+                // callers accidentally pass non-array data.
```
matthias-luger: remove AI comments

matthias-luger (changes requested):

components/MayorFlips/MayorFlips.tsx
Ekwav: because they are all in different pullrequests

Ekwav: @matthias-luger

PR #1446, opened 2025-11-19
added filter search info text

### LINKED ISSUE #1445: Add note that the search goes through options
At the bottom of the dropdown there should be a note that the filter search above should be used as it searches through all filter values as well
context:https://discord.com/channels/267680588666896385/267680588666896385/1440335215431254037

matthias-luger:
fixes #1445

### DIFF
--- a/components/ItemFilter/ItemFilter.tsx
+++ b/components/ItemFilter/ItemFilter.tsx
@@ -10,7 +10,7 @@ import HelpIcon from '@mui/icons-material/Help'
 import AddIcon from '@mui/icons-material/AddCircleOutline'
 import { camelCaseToSentenceCase, convertTagToName } from '../../utils/Formatter'
 import { FilterType, hasFlag } from '../FilterElement/FilterType'
-import { Typeahead, TypeaheadRef } from 'react-bootstrap-typeahead'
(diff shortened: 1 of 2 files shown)


components/ItemFilter/ItemFilter.tsx
```diff
@@ -399,75 +396,77 @@ function ItemFilter(props: Props) {
...
+                                                return (
+                                                    name.replace(/\s/g, '').includes(searchString) ||
+                                                    initials.includes(searchString) ||
+                                                    description.includes(searchString) ||
+                                                    matchingOptions.length > 0
+                                                )
+                                            }}
+                                            emptyLabel={props.emptyLabel || 'No matches found. Filters which would not show any results are hidden'}
+                                        ></Typeahead>
+                                        <p className={styles.filterHelpText}>You can search for filters both by their name and possible values (if available)</p>
```
Ekwav: To keep the UI clean and only let users searching for a specific filter to see this it should be a virtual option in the dropdown itself (selecting it not adding an actual filter)

Ekwav (changes requested):

components/ItemFilter/ItemFilter.tsx
matthias-luger: @Ekwav Opinion?
image

matthias-luger:
fixes #1445

PR #1450, opened 2025-11-24
Startpage revamp

Ekwav:


components/CoflCoins/CoflCoinAmountSelection.tsx
matthias-luger, line 1: I am not sure if this is the correct brach for this change, but i i gues i will still just let it be in here

components/Startpage/Startpage.module.css
```diff
@@ -73,3 +73,26 @@
...
 }
+
+.hover-effect {
+    transition: transform 0.2s ease-in-out, box-shadow 0.2s ease-in-out;
+}
+
+.hover-effect:hover {
+    transform: translateY(-5px);
+    box-shadow: 0 10px 20px rgba(0,0,0,0.2) !important;
+}
```
matthias-luger: why dont you call it hoverEffect here as well?

.gitignore
```diff
@@ -38,3 +38,4 @@ cypress/videos/*
 /test-results/
 /playwright-report/
 /playwright/.cache/
+data/
```
matthias-luger, line 41: whats this folder?

components/Startpage/Startpage.tsx
matthias-luger, line 1: I think the new Cards look worse than the old ones. If you want to keep them, please add a border and change the other Cards (new items) also to the new ones. Also I would prefer if the header and body would be seperated like in the old cards

matthias-luger, line 1: The new buttons "Visit Wiki" and "Get the mod" are extremely hard to see

components/Startpage/Startpage.module.css
Ekwav: forgot, did now

.gitignore
Ekwav: Reference data for writing guides

Issue #1451, opened 2025-11-25
Addition of sitemaps added conflicting middleware


matthias-luger:
The sitemaps PR added an additional middleware file, which is detected as a duplicate and might/is conflicting with the old one.

<img width="1327" height="382" alt="Image" src="https://github.com/user-attachments/assets/10f5c0a6-001a-4fd1-9800-f22686330e4c" />

PR #1454, opened 2025-12-03
remove duplicate middleware

matthias-luger:


matthias-luger:
@Ekwav It should work like before. Maybe you want to take a look.
Most notably the exported "config" in the old "middleware.ts". I didnt migrate it, as it seems like it is never used

matthias-luger:
fixes #1451

Ekwav (approved):
LGTM!

PR #1479, opened 2026-03-09
Add bazaar export dialog

### LINKED ISSUE #1477: Add bazaar export info page
context:https://discord.com/channels/909292333357469787/909298606022983690/1479244088233431100

Ekwav:
fixes https://github.com/Coflnet/hypixel-react/issues/1477

### DIFF
--- a/components/BazaarExportModal/BazaarExportModal.tsx
+++ b/components/BazaarExportModal/BazaarExportModal.tsx
@@ -3,6 +3,7 @@ import { useState } from 'react'
 import { Alert, Button, Form, Modal } from 'react-bootstrap'
 import api from '../../api/ApiHelper'
 import { hasHighEnoughPremium, PREMIUM_RANK } from '../../utils/PremiumTypeUtils'
+import { getGetApiBazaarItemTagExportUrl } from '../../api/_generated/skyApi'
 import styles from './BazaarExportModal.module.css'
 import GoogleSignIn from '../GoogleSignIn/GoogleSignIn'
 
@@ -45,7 +46,7 @@ function BazaarExportModal(props: Props) {
         )
     }
 
-    function handleEnter() {
+    function onModalEntered() {
         let loggedIn = !!(typeof window !== 'undefined' && (sessionStorage.getItem('googleId') || localStorage.getItem('googleId')))
         setIsLoggedIn(loggedIn)
 
@@ -59,15 +60,10 @@ function BazaarExportModal(props: Props) {
 
     function onLogin() {
         setIsLoggedIn(true)
-        handleEnter()
+        onModalEntered()
     }
 
     function handleModalHide() {
-        // Reset form state when modal closes
-        setFullOrderBook(false)
-        setStartDate('')
-        setEndDate('')
-        setIsExporting(false)
         props.onHide()
     }
 
@@ -100,6 +96,7 @@ function BazaarExportModal(props: Props) {
         const end = new Date(endDate)
         const diffTime = Math.abs(end.getTime() - start.getTime())
         const diffDays = Math.ceil(diffTime / (1000 * 60 * 60 * 24))
+        // 400 days is just a nice approximate for "more than a year" to warn about slow exports
         return diffDays > 400
     }
 
@@ -122,16 +119,13 @@ function BazaarExportModal(props: Props) {
 
         const googleToken = sessionStorage.getItem('googleId') ?? localStorage.getItem('googleId') ?? ''
 
-        const params = new URLSearchParams()
-        params.append('fullOrderBook', fullOrderBook.toString())
-        if (startDate) {
-            params.append('start', new Date(startDate).toISOString())
-        }
-        if (endDate) {
-            params.append('end', new Date(endDate).toISOString())
-        }
-
-        const url = `https://sky.coflnet.com/api/bazaar/${props.itemTag}/export?${params.toString()}`
+        // Using manual fetch instead of generated Tanstack query (e.g. useGetApiBazaarItemTagExport) 
+        // because the generated API attempts to parse the response as JSON, but this endpoint returns a binary ZIP blob.
+        const url = getGetApiBazaarItemTagExportUrl(props.itemTag, {
+            fullOrderBook: fullOrderBook,
+            start: startDate ? new Date(startDate).toISOString() : undefined,
+            end: endDate ? new Date(endDate).toISOString() : undefined
+        })
 
         fetch(url, {
             method: 'GET',
@@ -166,18 +160,18 @@ function BazaarExportModal(props: Props) {
     }
 
     return (
-        <Modal show={props.show} onHide={handleModalHide} onEntered={handleEnter} size="lg" centered>
+        <Modal show={props.show} onHide={handleModalHide} onEntered={onModalEntered} size="lg" centered>
             <Modal.Header closeButton>
                 <Modal.Title>Export Bazaar Data</Modal.Title>
             </Modal.Header>
             <Modal.Body>
                 {!isLoggedIn ? (
-                    <div style={{ textAlign: 'center', padding: '20px' }}>
+                    <div className={styles.centerPadded}>
                         <p>You need to be logged in to export bazaar data.</p>
                         <GoogleSignIn onAfterLogin={onLogin} />
                     </div>
                 ) : !hasLoadedPremium ? (
-                    <div style={{ textAlign: 'center', padding: '20px' }}>Loading...</div>
+                    <div className={styles.centerPadded}>Loading...</div>
                 ) : (
                     <>
                         {premiumLoadError && (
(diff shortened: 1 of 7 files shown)


components/Search/Search.tsx
```diff
@@ -56,7 +56,7 @@ function Search(props: Props) {
     let [results, setResults] = useState<SearchResultItem[]>([])
     let [isSearching, setIsSearching] = useState(false)
     let [noResultsFound, setNoResultsFound] = useState(false)
-    let [isSmall, setIsSmall] = useState(true)
+    let [isSmall, setIsSmall] = useState(false)
```
matthias-luger, line 59: Why this change?

components/OptionsMenu/OptionsMenu.tsx
```diff
@@ -84,6 +86,11 @@ function OptionsMenu(props: Props) {
                         <Button>{result.title}</Button>
                     </a>
                 ))}
+                {isItemPage && (props.selected as Item).bazaar && (
```
matthias-luger: the "isItemPage && (props.selected as Item).bazaar" check can be extracted into one variable that is set before the rendering logic like "isBazaarItem"

components/BazaarExportModal/BazaarExportModal.tsx
```diff
@@ -0,0 +1,352 @@
...
+                                    </ul>
+                                </div>
+                            </>
+                        )}
+                    </>
+                )}
+            </Modal.Body>
+            <Modal.Footer style={{ flexWrap: 'wrap', gap: '10px' }}>
+                {!isLoggedIn || (!hasPremium && !hasPremiumPlus) ? (
+                    <div style={{ width: '100%', marginBottom: '10px', fontSize: '0.85rem', color: '#666' }}>
```
matthias-luger: would be nice to extract all (or at least repeating stylings) into the css file

components/BazaarExportModal/BazaarExportModal.tsx
```diff
@@ -0,0 +1,352 @@
...
+        loadPremiumProducts()
+    }
+
+    function onLogin() {
+        setIsLoggedIn(true)
+        handleEnter()
+    }
+
+    function handleModalHide() {
+        // Reset form state when modal closes
```
matthias-luger: not necessary

components/BazaarExportModal/BazaarExportModal.tsx
```diff
@@ -0,0 +1,352 @@
...
+            () => {
+                setPremiumLoadError(true)
+                setHasPremium(false)
+                setHasPremiumPlus(false)
+                setHasLoadedPremium(true)
+            }
+        )
+    }
+
+    function handleEnter() {
```
matthias-luger: I would rename this function as I initially thought this was a handler for the keypress of "Enter", not if the Modal is enterd. Or add a comment to make this clear

components/BazaarExportModal/BazaarExportModal.tsx
```diff
@@ -0,0 +1,352 @@
...
+        return today.toISOString().split('T')[0]
+    }
+
+    function isDateRangeExceedsYear(): boolean {
+        if (!startDate || !endDate) return false
+        const start = new Date(startDate)
+        const end = new Date(endDate)
+        const diffTime = Math.abs(end.getTime() - start.getTime())
+        const diffDays = Math.ceil(diffTime / (1000 * 60 * 60 * 24))
+        return diffDays > 400
```
matthias-luger, line 100: why 400?

components/BazaarExportModal/BazaarExportModal.tsx
```diff
@@ -0,0 +1,352 @@
...
+                a.remove()
+                window.URL.revokeObjectURL(downloadUrl)
+                props.onHide()
+            })
+            .catch(err => {
+                alert('Export failed: ' + err.message)
+            })
+            .finally(() => {
+                setIsExporting(false)
+            })
```
matthias-luger, line 159: why do you fetch this manually? We now use our generated API via Tanstack Query

components/BazaarExportModal/BazaarExportModal.tsx
```diff
@@ -0,0 +1,352 @@
...
+                            <Alert variant="info">
+                                With Premium you can export up to the last <strong>180 days</strong> of data. Upgrade to <strong>Premium+</strong> for up to{' '}
+                                <strong>~6 years</strong> of history.
+                                <div style={{ marginTop: '10px' }}>
+                                    <Button href="/premium?tier=premium_plus" variant="warning">
+                                        Upgrade to Premium+
+                                    </Button>
+                                </div>
+                            </Alert>
+                        )}
```
matthias-luger, line 213: I think this doesnt look good. The orange button on the blue Alert feels wrong. Also text in the button being underlined also looks off (there is already a global class called disableLinkStyle to remove this if necessary)

components/BazaarExportModal/BazaarExportModal.tsx
```diff
@@ -0,0 +1,352 @@
...
+                        {isLoggedIn && !hasPremium && !hasPremiumPlus && !premiumLoadError && (
+                            'Premium subscription required to export bazaar data. Click "Get Premium" above to upgrade.'
+                        )}
+                    </div>
+                ) : (
+                    <div style={{ width: '100%', marginBottom: '10px', fontSize: '0.85rem', color: '#666' }}>
+                        By clicking Download, you accept the license agreement above.
+                    </div>
+                )}
+                <Button variant="outline-secondary" onClick={handleModalHide}>
```
matthias-luger: The cancel button is nearly invisible

components/BazaarExportModal/BazaarExportModal.tsx
```diff
@@ -0,0 +1,352 @@
...
+                        {!isLoggedIn && 'Please log in to export bazaar data.'}
+                        {isLoggedIn && !hasPremium && !hasPremiumPlus && premiumLoadError && (
+                            'Unable to verify premium status. Please retry.'
+                        )}
+                        {isLoggedIn && !hasPremium && !hasPremiumPlus && !premiumLoadError && (
+                            'Premium subscription required to export bazaar data. Click "Get Premium" above to upgrade.'
+                        )}
+                    </div>
+                ) : (
+                    <div style={{ width: '100%', marginBottom: '10px', fontSize: '0.85rem', color: '#666' }}>
```
matthias-luger: Maybe this is on purpose, but the text is extremely hard to read. I prefer the styling of the note block (styles.noteText). It is still grey, but still readable

matthias-luger (changes requested):

components/Search/Search.tsx
Ekwav: to avoid hydration issues on PC but I suppose it should not be in the bazaar dialog PR, reverted it

components/BazaarExportModal/BazaarExportModal.tsx
Ekwav: more than a year, this calculation is for the warning of slow exports and it should only show up if its actually slow, 400 days is just a nicer number than a precise 1 year and one month.
Clarrified that

components/OptionsMenu/OptionsMenu.tsx
Ekwav: did that

components/BazaarExportModal/BazaarExportModal.tsx
Ekwav: refactored

components/BazaarExportModal/BazaarExportModal.tsx
Ekwav: removed

components/BazaarExportModal/BazaarExportModal.tsx
Ekwav: renamed to onModalEntered

components/BazaarExportModal/BazaarExportModal.tsx
Ekwav: The generated Orval API code is hardcoded to assume and parse JSON responses, this is a zip download which does not work with the generated API

components/BazaarExportModal/BazaarExportModal.tsx
Ekwav: changed it to an info button

components/BazaarExportModal/BazaarExportModal.tsx
Ekwav: improved readability
image

components/BazaarExportModal/BazaarExportModal.tsx
Ekwav: bumped the contrast further
image

components/BazaarExportModal/BazaarExportModal.tsx
matthias-luger: You can keep it this way if it work i guess, but is the Swagger defined correctly? If it returns binary it should just return the binary string. At least thats what i would assume

matthias-luger (approved):

matthias-luger:
@Ekwav You can either take another look at the binary response or just merge it :)

components/BazaarExportModal/BazaarExportModal.tsx
Ekwav: it seems to be a general orval limitation https://github.com/orval-labs/orval/issues/869

PR #1508, opened 2026-05-09
Fix archived auction filter links

doej13367:
Fixes archived auction navigation and an outdated build guard.

- preserve active item filters when linking to archived auctions
- remove an obsolete Orval TypeScript suppression that breaks production builds
- refresh safe lockfile resolutions from the non-force npm audit fix

Confirmed with `npm run build`, `npx tsc --noEmit --pretty false`, `git diff --check`, and `npm audit --audit-level=high`.

### DIFF
--- a/app/about/page.tsx
+++ b/app/about/page.tsx
@@ -34,7 +34,7 @@ export default async function Page() {
                     <hr />
                     <h3>Item icons</h3>
                     <div>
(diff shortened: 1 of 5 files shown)


ptlthg:
This is a malicious PR, they changed  SkyCrypt links to a phishing site.

PR #1519, opened 2026-07-10
Add socket based live auction updates

Ekwav:


components/RecentAuctions/RecentAuctions.tsx
```diff
@@ -156,6 +199,8 @@ function RecentAuctions(props: Props) {
         }
 
         localStorage.setItem(RECENT_AUCTIONS_FETCH_TYPE_KEY, e.target.value)
+        // re-subscribe to live sold auctions with the newly selected fetch type
+        setFetchTypeVersion(v => v + 1)
```
matthias-luger: this is kinda bad practice. I would prefer it if you put the subscribe logic into its own hook. Inside there you can have a "resubscribe" method that the hook returns. This would eliminate the useEffect hook

matthias-luger (changes requested):

components/RecentAuctions/RecentAuctions.tsx
Ekwav: replaced with a hook

